PR #15, opened 2021-04-14
Finetune github action example

janlaan:
I propose 2 changes to the Github action example:

1. The `if (success() || failure())` check in this change saves the HTML report even if the analysis step fails, because that's when you need the report the most. The default behavior is to skip this step when the analysis step failed. 

2. Only run Sigrid CI on pull requests, not on pushes. With the current example, when you push to a PR, Sigrid CI runs twice, because that's both a push event as a PR-related event. I believe running these kind of checks on PR's is the most common scenario.


janlaan:
@dennis-sig FYI

PR #16, opened 2021-04-15
Retry with friendly error message on HTTP 5xx server.

dennis-sig:


sigridci/sigridci.py
```diff
@@ -99,15 +99,22 @@ def fetchAnalysisResults(self, analysisId):
...
             time.sleep(self.POLL_INTERVAL)
             
         log("Analysis failed: waiting for analysis results took too long")
         sys.exit(1)
         
+    def processHttpError(self, e):
+        if e.code == 404:
+            log("Analysis results not yet available")
+        elif e.code >= 500:
+            log("HTTP status {e.code} when connecting to Sigrid, retrying")
```
tukraus, line 114: I think this should be an f-string ?

test/sigridcitest.py
```diff
@@ -140,6 +141,15 @@ def testForceLowerCaseForCustomerAndSystemName(self):
     def testFeedbackTemplateOnlyContainsAsciiCharacters(self):
         with open("sigridci/sigridci-feedback-template.html", mode="r", encoding="ascii") as templateRef:
             template = templateRef.read()
+            
+    def testOnlyThrowExceptionForClientError(self):
+        args = types.SimpleNamespace(partner="sig", customer="Aap", system="NOOT", sigridurl="")
+        apiClient = SigridApiClient(args)
+        apiClient.processHttpError(urllib.error.HTTPError("http://www.sig.eu", 404, "", {}, None))
+        apiClient.processHttpError(urllib.error.HTTPError("http://www.sig.eu", 502, "", {}, None))
```
tukraus, line 149: Maybe split this in 2 tests where one expects the Exception (client error) and the other (server error) does not?

tukraus (changes requested):

janlaan:
This fixes the issue when the error occurs during `fetchAnalysisResults`. It does not address the scenario where the HTTP call in`submitUpload` fails with a 4xx or 5xx HTTP error.

dennis-sig:
@tukraus @janlaan Could you have another look? Error message in the various situations should be more aligned now.

janlaan:
> @tukraus @janlaan Could you have another look? Error message in the various situations should be more aligned now.

LGTM

tukraus (approved):

PR #32, opened 2021-06-24
Fix typos in pathprefix and clarify readme

janlaan:
This PR makes sure that the parameter is referred to as `pathprefix` everywhere.
It also adds some more clarification to the explanation, I thought it was a bit ambiguous.


janlaan:
@dennis-sig

dennis-sig:
Thanks @janlaan !

PR #34, opened 2021-06-30
Add option to show upload contents, and add warning when upload appears to be incomplete

dennis-sig:


dennis-sig:
@Geertien My initial commit failed because I used the `list[str]` hint, which was only introduced in Python 3.9. To keep compatibility with Python 3.8 and older I changed it to the old way (see https://stackoverflow.com/questions/57690749/how-to-declare-an-array-or-a-list-in-a-python-dataclass)

MichielCuijpers:
this will help users identify why they have an empty upload to send to the SigridCi endpoint.

Geertien (approved):
Looks good.

Nice addition of the `UploadOptions` class.

PR #42, opened 2021-08-03
Document --publish option for GitLab.

dennis-sig:


faq.md
```diff
@@ -33,6 +35,16 @@ Sigrid CI: Frequently Asked Questions
 
 ## Usage questions
 
+### Do you support pull request integration?
+
+Yes! There are basically two usage scenarios for Sigrid CI, and you would typically use them both.
+
+The first Sigrid CI use case is to *publish* your project's source code to Sigrid, which makes the analysis results accessible on [https://sigrid-says.com](https://sigrid-says.com). You would typically use this for your project's main branch, which can be either `main`, `master`, or something project-specific.
```
Geertien, line 42: I'd get rid of "either" here.

```suggestion
The first Sigrid CI use case is to *publish* your project's source code to Sigrid, which makes the analysis results accessible on [https://sigrid-says.com](https://sigrid-says.com). You would typically use this for your project's main branch, which can be `main`, `master`, or something project-specific.
```

faq.md
```diff
@@ -33,6 +35,16 @@ Sigrid CI: Frequently Asked Questions
...
+
+The second use case is to use Sigrid CI for feedback during code reviews on pull requests. In these situations, you would typically *not* publish the results to Sigrid, since the code review is usually been author and reviewer and does not require the entire team to see the results.
```
Geertien, line 44: > since the code review is usually been author and reviewer and does not require the entire team to see the results.
I'm not sure what you mean here. Should "been" be "between"? Then the sentence makes sense to me.

On another note, I would say another important reason is that Sigrid CI compares to the code as published to Sigrid, so in general you would not want to publish code that has not yet been merged into your main branch. Would that be good to add here?

faq.md
```diff
@@ -47,6 +59,14 @@ You will find the system name in the url of the monitor in Sigrid. The structure
 
 Please send and email to support@softwareimprovementgroup.com and mention your name, your company and the system name(s) that you need a token for. We will send the account via mail and the token via sms.
 
+### What target quality should we use?
+
+Avoid setting an unreasonably high target quality level. While it seems appealing to be ambitious, this can be demotivating or frustrating to people, as it's not always possible to achieve such a high level in every single change. .
```
Geertien, line 64:
```suggestion
Avoid setting an unreasonably high target quality level. While it seems appealing to be ambitious, this can be demotivating or frustrating to people, as it's not always possible to achieve such a high level in every single change.
```

Geertien (approved):
I made a few (small) suggestions for you to look at. Otherwise it looks good.

faq.md
dennis-sig: We do explain that point https://github.com/Software-Improvement-Group/sigridci/blob/main/faq.md#what-branch-or-commit-does-sigrid-ci-compare-against. I'm hesitant to repeat the entire argument in the platform-specific instructions. As a middle ground, let's include a very brief description in these bullets.

dennis-sig: Oh, and the "been/between" was indeed a typo :)

PR #44, opened 2021-09-14
Update bitbucket-pipelines.md

bitisuvanje:
Better explained, in the example code snippet, that only the Sigrid CI step needs Python 3.7 or higher to run, and any other image can be used as build environment.


MichielCuijpers:
thanks @bitisuvanje  We will look at you PR. Best regards Michiel Cuijpers of SIG.

dennis-sig:
@bitisuvanje You're correct, the Sigrid CI step can indeed use a different image from the rest of the pipeline. Thanks a lot for the PR!

PR #47, opened 2021-09-30
Explain why --publish and --pathprefix cannot be used simultaneously.

dennis-sig:


MichielCuijpers:
thanks this makes sense.

PR #56, opened 2021-10-27
Support configurable thresholds per system property

dennis-sig:


client-script-usage.md
```diff
@@ -23,6 +33,25 @@ Once the `sigridci.py` script is available within your CI environment, you can c
...
 
+## Specifying a quality target
+
+Sigrid CI compares the quality of the new/changed code against the configured target quality level. The simplest and recommended way to configure the target is by using the `--targetquality` command line argument. This will check the overall quality maintainability rating against the target. 
+
+The advanced approach requires a configuration file named `sigrid.yaml` located in the root of the repository. This approach allows you to specify a target quality level for both the overall maintainability level and every system property. You can use a different target for different metrics, so this allows you to both have an overall target, but also be more strict (or more lenient) for some of the underlying system properties. Using the configuration file provides more flexibility and more control, but it also makes the feedback more complicated. In general, it is recommended to use the overall target quality level, and only start defining specific thresholds for specific system properties when there are structural quality issues that cannot be addressed otherwise.
```
bruno-oliveira, line 40: Maybe we can re-word as: 
"You can use a different target for different metrics..." -----> "You can use different targets per metric"?

sigridci/sigridci.py
```diff
@@ -45,6 +46,32 @@ class UploadOptions:
...
+            log(f"Loading target quality level from configuration file {configFile}")
+            # We can't use pyyaml because PIP is not available in the some of the
+            # very diverse set of customer environments where Sigrid CI is used.
+            targetPattern = re.compile("(" + "|".join(Report.METRICS) + "):\s*([\d\.]+)", re.IGNORECASE)
+            for line in open(configFile, "r"):
+                match = targetPattern.match(line.strip())
+                if match:
+                    self.ratings[match.group(1).upper()] = float(match.group(2))
+    
+    def isPassed(self, feedback, metric):
```
bruno-oliveira, line 68: Would it be an idea to rename this method to something more descriptive like: `meetsTargetQualityForMetric`?

sigridci/sigridci.py
```diff
@@ -45,6 +46,32 @@ class UploadOptions:
...
+        value = feedback["newCodeRatings"].get(metric, None)
+        targetRating = self.ratings.get(metric, None)
+        return value == None or targetRating == None or value >= targetRating
+        
+    def isOverallPassed(self, feedback):
```
bruno-oliveira, line 73: This could be something like: `meetsOverallQualityTarget`?

sigridci/sigridci.py
dennis-sig: You're right, will rename this one and the other method name you suggested.

client-script-usage.md
dennis-sig: I suggest a middle ground in "You can use different targets for different metrics". I'm afraid some people coudl misinterpret "per metric".

PR #61, opened 2021-12-02
Clarify target quality analysis on publish step

jsarvis:
To mirror other pipeline examples, in Bitbucket pipeline example publish step specify the target quality instead the implicit default to make it clear this example is both analyzing with a target quality and publishing, and thus could display text indicating the target quality was not met if the analysis fails to meet the target quality.

### DIFF
--- a/bitbucket-pipelines.md
+++ b/bitbucket-pipelines.md
@@ -55,7 +55,7 @@ pipelines:
           image: python:3.9 # The client scripts for Sigrid CI are based on Python.
           script:
             - "git clone https://github.com/Software-Improvement-Group/sigridci.git sigridci"
-            - "./sigridci/sigridci/sigridci.py --customer examplecustomername --system examplesystemname --source . --publish"
+            - "./sigridci/sigridci/sigridci.py --customer examplecustomername --system examplesystemname --source . --targetquality 3.5 --publish"
   pull-requests:
     - step:
         name: Sigrid CI


jsarvis:
@MichielCuijpers 
Otherwise, I suggest changing the publish to --publishonly without a target quality, to just publish a baseline.
Especially if pull requests typically target the branch identified for publishing a baseline, as an identical post-merge commit would have already been analyzed in the pull request.

If a different target quality was expected on the branch, then it may make sense to also analyze there to block subsequent build/deploy steps, but that should be called out as a strategy.

Either way, the current example doesn't make it clear that publish will also analyze and will use default target quality when not specified.

MichielCuijpers:
Hi @jsarvis during the --publish(only) steps we will not generate an error exitcode to fail the build. Also not in the case when the quality of the repo is lower than the targetquality. We made this PR to make that more clear in the script https://github.com/Software-Improvement-Group/sigridci/pull/59/files

dennis-sig:
@jsarvis Thanks for creating the pull request, but @MichielCuijpers is correct. You can basically use Sigrid CI in two ways: as a quality gate for pull requests, and to publish your project to the Sigrid dashboard. The `—targetquality` option only applies to the first scenario, but not for the second. So the value of this argument will never fail your pipeline when you're publishing source code .

So we propose to not do this change, as it would be confusing to document this option for a scenario where it doesn't really apply. However, you're totally right that we *should* explain a lot better when and how this `—targetquality` option is used, and I would propose that we make that change instead.

jsarvis:
Ah, yes. Thank you for highlighting that line @MichielCuijpers 
So it won't return an error exitcode, but for `--publish` it will display text that the target quality was not met based on the implicit target quality of 3.5 (if it didn't meet that quality during the analysis).
That is much less dire, but could still be confusing.
Either the publish steps could be changed to use publishonly in the examples, or the strategy of which target quality is used on the same branch/build step that uses publish would be helpful.
This example using publish (instead of publishonly) and not specifying a target quality would at least benefit from calling out that the default quality will be used for an analysis on that step, which will display results, but not fail the build.
I only see from the python source the clarity of when an error exitcode will be returned. Perhaps also a note would be helpful for the publish argument stating that an analysis will be displayed but no errors returned from not meeting the target quality.

PR #62, opened 2021-12-13
Only use ASCII art color when metric has an explicit target, to avoid misunderstandings

dennis-sig:


Geertien (approved):
Looks good to me. Although I've left one comment on the code. You might want to consider making a change based on that, but if you don't agree I'm also fine with that and see no problem with merging.

PR #63, opened 2022-01-10
Exclude $tf directory by default.

dennis-sig:


MichielCuijpers (approved):
Yes lets remove these TFS files from the SigridCI upload

PR #68, opened 2022-01-19
Show overall ratings when using —publish

dennis-sig:


Geertien (approved):
The changes look good to me.

Just to make sure, I wanted to check the behaviour when no baseline snapshot is available. If I understand correctly in this case we will have "Baseline on N/A" in the header and "(N/A)" for every score (metrics and Maintainability). I considered some alternatives like having "Baseline unavailable" or something similar in the header, but I think the current implementation is clearest in the end.

I also considered completely removing the baseline column and/or displaying the overall rating, if there is no baseline available. However, I think the lack of continuity will be more confusing, so I think the current choice is best.

I wanted to share the above considerations, but if my assertions there were correct and you agree with my conclusions, I fully approve of the merging these changes as they are.

Also some nice refactoring in there!

dennis-sig:
Fair point, the UX when we no longer have the baseline results is not great. Considering the small number of occurrences (< 1%), we'll leave that for a follow-up ticket.

PR #71, opened 2022-02-04
Also validate the system name client-side to give early feedback if it's invalid

dennis-sig:


sigridci/sigridci.py
```diff
@@ -31,6 +31,7 @@
 
 
 LOG_HISTORY = []
+SYSTEM_NAME_PATTERN = re.compile("[a-z0-9-]{2,}", re.IGNORECASE)
```
patveck, line 34: A system name needs to be lower case. In addition, it cannot start with a dash, hence my suggestion.
```suggestion
SYSTEM_NAME_PATTERN = re.compile("[a-z0-9][a-z0-9-]+")
```

sigridci/sigridci.py
```diff
@@ -533,6 +534,10 @@ def run(self, apiClient, options, target, reports):
     if args.publish and len(args.pathprefix) > 0:
         print("You cannot use both --publish and --pathprefix at the same time, refer to the documentation for details")
         sys.exit(1)
+        
+    if not SYSTEM_NAME_PATTERN.match(args.system):
+        print(f"System name '{args.system}' is invalid, system names can only contain letters, numbers, and hyphens")
```
patveck, line 539:
```suggestion
        print(f"System name '{args.system}' is invalid, system names can only contain lowercase letters, numbers, and hyphens, and cannot start with a hyphen")
```

test/test_sigridci.py
```diff
@@ -125,9 +125,18 @@ def testLoadTargetsFromConfigFile(self):
...
         self.assertEqual(target.ratings.get("UNIT_SIZE", None), None)
+    
+    def validateSystemNameAccordingToRules(self):
+        self.assertTrue(SYSTEM_NAME_PATTERN.match("aap"))
+        self.assertTrue(SYSTEM_NAME_PATTERN.match("aap-noot"))
+        self.assertTrue(SYSTEM_NAME_PATTERN.match("aap123"))
+        
+        self.assertFalse(SYSTEM_NAME_PATTERN.match("aap_noot"))
+        self.assertFalse(SYSTEM_NAME_PATTERN.match("a"))
+        self.assertFalse(SYSTEM_NAME_PATTERN.match("$$$"))
```
patveck, line 136:
```suggestion
        self.assertFalse(SYSTEM_NAME_PATTERN.match("$$$"))
        self.assertFalse(SYSTEM_NAME_PATTERN.match("-aap"))
        self.assertFalse(SYSTEM_NAME_PATTERN.match("aAP"))
```

sigridci/sigridci.py
dennis-sig: @patveck We do a `.lower()` before using the customer and system name. I agree with the rest of the change, but if we disallow lowercase system names the script will break for peope that rely on the current implicit conversion to lowercase.

patveck (approved):

PR #77, opened 2022-03-01
Include and use pem-file

anoordover:
Include and use pem-file with root, intermediate and site certificate to trust sigrid-says.com


MichielCuijpers:
hi @anoordover  as discussed thanks for your PR. At this time we do not want overide our default behaviour and to include the pem files for all our customers. We do understand that on your server our https certificate is not recognised as trusted. 
It might be unrelated but the following could be something to check on your end https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/

MichielCuijpers:
hi @anoordover  please see my email with solutions. closing this PR

PR #79, opened 2022-03-02
removed some plurals

MichielCuijpers:


MichielCuijpers:
@dennis-sig noticed some plurals

PR #87, opened 2022-03-23
Add warning message for old tokens.

dennis-sig:


MichielCuijpers:
Thanks for adding this warning. We aim at deactivating the old tokens on April 1st.

PR #90, opened 2022-04-06
Reject upload of ZIP files with 0 entries in them.

dennis-sig:


sigridci/sigridci.py
```diff
@@ -243,31 +243,34 @@ def __init__(self, options):
...
         zipFile = zipfile.ZipFile(outputFile, "w", zipfile.ZIP_DEFLATED)
+        zipContents = []
         
         for root, dirs, files in os.walk(sourceDir):
             for file in sorted(files):
                 filePath = os.path.join(root, file)
                 if file != outputFile and not self.isExcluded(filePath):
                     relativePath = os.path.relpath(os.path.join(root, file), sourceDir)
                     uploadPath = self.getUploadFilePath(relativePath)
+                    zipContents.append(uploadPath)
```
Geertien, line 254: Using a list seems a bit overkill here. Just using a boolean `zipHasContent` which you set to `false` at 246 and `true` here does the trick just as well right? Or do you expect we want to do more with that list of paths later?

sigridci/sigridci.py
```diff
@@ -243,31 +243,34 @@ def __init__(self, options):
...
+    def checkUploadContents(self, outputFile, zipContents):
         uploadSizeBytes = os.path.getsize(outputFile)
         uploadSizeMB = max(round(uploadSizeBytes / 1024 / 1024), 1)
         log(f"Upload size is {uploadSizeMB} MB")
         
         if uploadSizeMB > self.MAX_UPLOAD_SIZE_MB:
             raise Exception(f"Upload exceeds maximum size of {self.MAX_UPLOAD_SIZE_MB} MB")
             
-        if uploadSizeBytes < 50000:
-            log("Warning: Upload is very small, source directory might not contain all source code")
```
Geertien, line 270: Maybe we still want to give this warning as `elif` after checking whether there are contents at all?

Geertien (changes requested):
Looks good, but I have some minor suggestions and discussion points I'd like you to look at.

sigridci/sigridci.py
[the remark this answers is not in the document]
dennis-sig: Correct. `os.walk` is weird in that it returns `root, dirs, files` as different lists, which is different from Java where you just get everything. Since we only iterate `files`, this indeed does not get called for directories.

Geertien (changes requested):
The changes look good. Something I missed yesterday though, is that I think it's good to also explicitly test this new rejecting/failing behaviour. Probably in `test_system_upload_packer` something like:
```
with self.assertRaises(SystemExit) as cm:
    uploadPacker.prepareUpload(sourceDir, outputFile)

self.assertEqual(cm.exception.code, 1)
```

Geertien (approved):
Looks good to me 🙂

PR #91, opened 2022-04-07
Retry and stream upload to S3.

dennis-sig:


sigridci/sigridci.py
```diff
@@ -156,14 +154,24 @@ def getRequestUploadPath(self, systemExists):
...
         
     def uploadBinaryFile(self, url, upload):
+        for attempt in range(self.RETRY_ATTEMPTS):
+            try:
+                return self.attemptUpload(url, upload)
+            except urllib.error.HTTPError as e:
+                log("Retrying upload")
+                time.sleep(self.POLL_INTERVAL)
+        
+        log("Sigrid is currently unavailable")
```
Geertien, line 164: Do you really want this to be `"Sigrid is currently unavailable"` and not the original `"Uploading file failed"`, or even better probably `f'Uploading file failed after {self.RETRY_ATTEMPTS} attempts'`? (As this is also/mainly added for the case where Sigrid is available but the upload times-out for some other reason.)

dennis-sig:
@Geertien I'm aware that the repetition between retry attempts is bad. The problem is that various end points have their own behavior. For example, the system exists end point returns 404 as a valid response status. These differences make it a bit hard to generalize.

sigridci/sigridci.py
```diff
@@ -156,14 +154,24 @@ def getRequestUploadPath(self, systemExists):
...
         with open(upload, "rb") as uploadRef:
-            uploadRequest = urllib.request.Request(url, data=uploadRef.read())
+            uploadRequest = urllib.request.Request(url, data=uploadRef)
             uploadRequest.method = "PUT"
             uploadRequest.add_header("Content-Type", "application/zip")
             uploadRequest.add_header("Content-Length", "%d" % os.path.getsize(upload))
             uploadRequest.add_header("x-amz-server-side-encryption", "AES256")
-            uploadResponse = urllib.request.urlopen(uploadRequest)
-            return uploadResponse.status in [200, 201, 202]
+            return urllib.request.urlopen(uploadRequest)
```
Geertien, line 174: There's no more need to return a result here and subsequently in `uploadBinaryFile` anymore right, as we no longer check the status, but `sys.exit(1)` instead when we run out of attempts right?

sigridci/sigridci.py
```diff
@@ -127,10 +127,8 @@ def submitUpload(self, options, systemExists):
         analysisId = uploadLocation["ciRunId"]
         log(f"Sigrid CI analysis ID: {analysisId}")
         log("Publishing upload" if self.publish else "Submitting upload")
-
-        if not self.uploadBinaryFile(uploadUrl, upload):
-            raise Exception("Uploading file failed")
-            
+        self.uploadBinaryFile(uploadUrl, upload)
+
```
Geertien, line 131: Do we want to add an "Upload successful" (or similar) log message for completeness?

test/test_sigridci_runner.py
```diff
@@ -158,6 +158,60 @@ def testExitWithMessageIfUploadEmpty(self):
         runner = SigridCiRunner()
         with self.assertRaises(SystemExit):
             runner.run(apiClient, options, target, [])
+            
+    def testRetryIfUploadFailsTheFirstTime(self):
```
Geertien, line 162: Nice testing 😄

Geertien (changes requested):

sigridci/sigridci.py
dennis-sig: Indeed, we could drop the return value.

sigridci/sigridci.py
dennis-sig: Good one, will change.

Geertien (approved):
Very nice. Looks good to me 🙂

PR #101, opened 2022-05-24
Update faq.md

MichielCuijpers:


docs/faq.md
```diff
@@ -160,8 +160,8 @@ So which one should you use? That depends on your development process. If you us
 ### What to do when the script does not work?
 
 The Sigrid CI Python script is currently under active development, which means that its subject to change, and namely, its dependencies are subject to change as we add new functionality.
-Sigrid CI requires Python 3.7 or higher.
-When not using Docker to run the script, this requires that the dependencies in the environment in which the script is being ran are in sync with the actual code, and, in order to do that, you can run either `pip install -r requirement.txt` or `pipenv install`, when using Python virtual environments.
+Sigrid CI requires Python 3.7 or higher. 
+When not using Docker to run the script, this requires that the dependencies in the environment in which the script is being ran are in sync with the actual code, and, in order to do that, you can run either `pip install -r requirement.txt` or `pipenv install`, when using Python virtual environments. In some cases we see that both 3.6 and 3.7 get installed in the runner, please make sure that 3.7 is the default.
```
Geertien, line 164: I think the added example is too specific, it's not necessarily 3.6 and 3.7 but can happen with any version combination. I would suggest the following, but any explanation that gives that nuance is fine with me.

Moreover I think we should place it before the information about pip install, as it relates more to the required version, so I've also turned that around in my suggestion.

```suggestion
In some cases multiple python versions are installed in the environment and the wrong one (below 3.7) may be selected automatically. If that is the case, please make sure that the highest version is set to be the default.
When not using Docker to run the script, this requires that the dependencies in the environment in which the script is being ran are in sync with the actual code, and, in order to do that, you can run either `pip install -r requirement.txt` or `pipenv install`, when using Python virtual environments.
```

MichielCuijpers: I agree, your version is better

Geertien (approved):
Looks good to me now 🙂

PR #105, opened 2022-07-04
Instructions for bulk-uploading projects to Sigrid.

dennis-sig:


nicorikken:
I tackled this goal differently:

- All in Python
- Use pygithub to request all repo's
- Handle all repo's sequentially in a for-loop:
  - Check if the code is relevant based on GitHub repo object properties using custom conditions
  - Git checkout into a temporary directory
  - Determine metadata from repo contents and generate a sigrid-friendly system name
  - Upload by calling the `sigrid.py` from a subprocess
- I optimized it to check how current the code in Sigrid is, and to skip if a recent upload already took place

Compared to the approach in this PR it allows you to inject logic to select the relevant code and can prevent unnecessary uploads.

The sequential setup makes it a bit slow, but that comes with the benefit of reducing the load on the side of Sigrid.

dennis-sig:
Closing this pull request because people are now able to figure this out without the need for special documentation.

PR #112, opened 2022-08-05
Publish docker image

patveck:
First step to publish an official SIG Sigrid CI Docker image on Docker Hub


dennis-sig:
@patveck Very small request: Could you change the metadata to have the display name "Sigrid | Software Assurance Platform"? The pipe is part Sigrid's typesetting, so you'll do our marketing colleagues a favor by keeping the branding consistent.

PR #115, opened 2022-08-12
Add FAQ entry for UTF-8 issue.

dennis-sig:


valdore86:
Can we provide an example of adding an environment variable to the workflow to show how to add `export PYTHONIOENCODING=utf8`

Like so?
```
name: "Run Sigrid CI" 
        env:
          SIGRID_CI_TOKEN: "${{ secrets.SIGRID_CI_TOKEN }}"
          PYTHONIOENCODING=utf8
```

dennis-sig:
@valdore86 Good idea, will add. GitLab Actions and GitLab default to UTF-8 anyway, but I'll add it for Jenkins and Azure.

dennis-sig:
@valdore86 Are you OK with these additions?

PR #118, opened 2022-08-17
Explicitly define TLS 1.2 for the connection with S3.

dennis-sig:


sigridci/sigridci.py
```diff
@@ -175,7 +175,7 @@ def attemptUpload(self, url, upload):
             uploadRequest.add_header("Content-Type", "application/zip")
             uploadRequest.add_header("Content-Length", "%d" % os.path.getsize(upload))
             uploadRequest.add_header("x-amz-server-side-encryption", "AES256")
-            urllib.request.urlopen(uploadRequest, context=SSLContext(PROTOCOL_TLS_SERVER))
+            urllib.request.urlopen(uploadRequest, context=SSLContext(PROTOCOL_TLS))
```
tukraus: The `PROTOCOL_TLS` is deprecated as of Python 3.10: you can use `PROTOCOL_TLS_CLIENT` instead
https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS

dennis-sig: I noticed, but with `PROTOCOL_TLS_CLIENT` it doesn't work, so the behavior is not identical. I suggest we leave this branch open to see if this actually solves the issue, but we don't need to proactively merge this.

dennis-sig:
Closing this pull request since the issue appears to have been specific to that one environment, and has not been reported since.

PR #121, opened 2022-08-19
Update system name length and regex requirements.

dennis-sig:


sigridci/sigridci.py
```diff
@@ -608,15 +613,15 @@ def displayMetadata(self, apiClient):
...
     log("Starting Sigrid CI")
+    
     options = UploadOptions(args.source, args.exclude.split(","), args.include_history, args.pathprefix, args.showupload, args.publishonly)
     target = TargetQuality(options.readScopeFile() or "", args.targetquality)
     apiClient = SigridApiClient(args)
     reports = [TextReport(), StaticHtmlReport(), JUnitFormatReport(), ExitCodeReport()]
 
     runner = SigridCiRunner()
+    if not runner.isValidSystemName(args.system):
+        print(f"Invalid system name, system name should match {SYSTEM_NAME_PATTERN} and within {SYSTEM_NAME_LENGTH} characters")
```
Geertien, line 625: Gramatically I think you should add "be" between "and" and "within", however if I'm not mistaken `{SYSTEM_NAME_LENGTH}` is printed as "range(2, 65)" which doesn't read very nicely. Also directly using `{SYSTEM_NAME_PATTERN}` gives "re.compile("^[a-z0-9]+(-[a-z0-9]+)*$", re.IGNORECASE)" and I think it's nicer to stick to just the pattern. I'd go for something like:

```suggestion
        print(f"Invalid system name, system name should match '{SYSTEM_NAME_PATTERN.pattern}' "
              f"and be {SYSTEM_NAME_LENGTH.start} to {SYSTEM_NAME_LENGTH.stop} characters long (inclusive).")
```

Geertien (changes requested):
Looks good to me apart from the 1 suggestion.

sigridci/sigridci.py
dennis-sig: Whoops, you're totally right. Those fields used to be a string and integer originally, but that way of printing indeed isn't very friendly.

dennis-sig:
Cool! I already informed our support team, so that makes us good to go.

PR #122, opened 2022-08-19
Update mendix-github-actions.md

MichielCuijpers:
We no longer need to set the Gitlab environment CI_PROJECT_DIR: "." for clients that use GitHub.


MichielCuijpers:
Need to wait with this merge until @dessens0 updates of the MendixPreProcessor Docker

PR #123, opened 2022-09-04
Make the SigrigPublish job consistent

gsmeets:
The docker example uses --publishonly, whereas the CLI script example uses --publish.
I would suggest keeping this consistent.


dennis-sig:
@gsmeets Thanks for fixing this inconsistency. In general we use `—publish` in a workflow without pull requests, such as Jenkins. The examples for Azure DevOps should have indeed used `—publishonly` from the start.

PR #134, opened 2022-10-31
Add FAQ entry for Git warning about directory ownership.

dennis-sig:


MichielCuijpers:
I think it is safe to trust this PR ;-)

PR #137, opened 2022-11-04
Switch from Alpine to Debian

patveck:


dessens0:
Why wouldn't we just build 2 images and tag them accordingly?

Dockerfile.hub
```diff
@@ -11,12 +11,13 @@ LABEL org.opencontainers.image.source="https://github.com/Software-Improvement-G
...
+RUN set -eux; \
+	apt-get update; \
+	apt-get install -y --no-install-recommends git; \
+	rm -rf /var/lib/apt/lists/*
 
 COPY --link sigridci /sigridci
 
 ENV PATH /sigridci:$PATH
 
-USER sigridci
```
dessens0, line 20: This would mean we run as root, right?

patveck: Yes. I'm asking around internally to see if there's any other way, but AFAIK the fifth requirement leaves no other way.

dessens0 (approved):

PR #138, opened 2022-11-06
Upgrade Alpine in Dockerfile

patveck:


patveck:
@dennis-sig Pushed to `softwareimprovementgroup/sigridci:staging`. Docker Hub reports no vulnerabilities, don't know whether that's because of my change or because the base image we use got refreshed.

PR #141, opened 2022-11-21
Make sure Git is available in GitHub Actions Dockerfile

dennis-sig:
We recently added Git to the Dockerfile that is published to DockerHub. However, we also need to add it to the GitHub Actions DockerFile, since that is also based on Alpine. This will only be noticed by the subset of customers that use GitHub Marketplace *and* enable the `—include-history` option for Architecture Quality, which is why we didn't notice this before.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,7 @@
 FROM python:3.10-alpine
 
+RUN apk --no-cache -U upgrade && \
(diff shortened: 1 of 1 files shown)


patveck (approved):
Same pattern we use for the Docker Hub dockerfile, so this seems perfectly OK to me!

PR #143, opened 2022-11-25
Also support metadata via environment variables or GitHub Action parameters

dennis-sig:
.

### DIFF
--- a/docs/metadata.md
+++ b/docs/metadata.md
@@ -3,7 +3,7 @@ Adding business context to a system using metadata
 
 Adding business context to Sigrid makes it easier to interpret the results. For example, a below-market average maintainability rating of 2 stars might seem like a problem, but this depends entirely on the context. If the system is no longer actively maintained and will be decommissioned by the end of the year (i.e., its lifecycle phase is end-of-life, or EOL for short), such a rating might be perfectly acceptable. But if that system is new, uses a modern technology, and is business critical, such a rating would be considered a red flag. In both cases the technical conclusion is identical, it's the context that determines the urgency.
 
-This context information is called *metadata* in Sigrid. Adding metadata can be done in 3 different ways:
+This context information is called *metadata* in Sigrid. Adding metadata can be done in 4 different ways:
 
 ### Option 1: Adding metadata in Sigrid
 
@@ -37,7 +37,6 @@ metadata:
   applicationType: ANALYTICAL
   externalId: ab12345
   isDevelopmentOnly: false
-  deploymentType: INTERNAL
   remark: "Some notes"
 ```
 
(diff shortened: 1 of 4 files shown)


docs/metadata.md
```diff
@@ -37,7 +37,6 @@ metadata:
   applicationType: ANALYTICAL
   externalId: ab12345
   isDevelopmentOnly: false
-  deploymentType: INTERNAL
```
patveck, line 40: Why are you deleting this line?

docs/metadata.md
```diff
@@ -49,6 +48,15 @@ metadata:
 
 The contents of the YAML file will be used to update the metadata whenever you publish your system to Sigrid. If you run Sigrid CI *without* publishing, i.e. when you run it for a branch or pull request, the metadata does *not* get updated. This ensures that publishing code and publishing metadata behave in a consistent way.
 
+### Option 4: Adding metadata via Sigrid CI parameters
+
+Metadata can also be configured by passing the metadata values as parameters when running Sigrid CI. This will dynamically generate the YAML file from option 3, but does not require you to commit this YAML file to your repository. 
+
+- For all platforms, you can define environment variables to the Sigrid CI run. The name of the environment variables is the lowercase version of the metadata fields listed above. For example, defining the `applicationType` field in the YAML is equivalent to defining an environment variable named `applicationtype`.
```
patveck, line 55:
```suggestion
- For all platforms, you can define environment variables to the Sigrid CI run. The name of the environment variables is the lowercase version of the metadata fields listed above. For example, defining the `applicationType` field in the YAML is equivalent to defining an environment variable named `applicationtype` as a Sigrid CI parameter.
```

sigridci/sigridci.py
```diff
@@ -606,9 +620,24 @@ def displayMetadata(self, apiClient, options):
             for key, value in apiClient.fetchMetadata().items():
                 if value:
                     print(f"    {key}:".ljust(20) + str(value))
+                    
+    def prepareMetadata(self, options):
+        metadata = {field: os.environ.get(field.lower(), "") for field in self.METADATA_FIELDS}
+        metadata = {field: value for field, value in metadata.items() if value and len(value) > 0}
```
patveck, line 626: Is this common Python idiom? We first initialize `metadata` in line 625 and then apply some filter-like construct in this line. Would it be possible to do this all at once, a bit like how you'd do this with Java's streams?

docs/metadata.md
dennis-sig: It was a duplicate, `deploymentType` was already mentioned on line 36.

sigridci/sigridci.py
dennis-sig: You can, but then the line would be very long. The downside is that you would need to repeat the logic both on the left of the dict comprehension and on the filter:

    metadata = {field: os.environ.get(field.lower(), "") for field in self.METADATA_FIELDS if os.environ.get(field.lower(), "")}

Alternatively, this would probably be nicer:

    getMetadataValue = lambda field: os.environ.get(field.lower(), "")
    metadata = {field: getMetadataValue(field) for field in self.METADATA_FIELDS if getMetadataValue(field)}

test/test_sigridci_runner.py
```diff
@@ -403,6 +406,66 @@ def testDoNotValidateMetadataFileIfNotPresent(self):
...
+        os.environ["externalid"] = "1234"
+        
+        runner = SigridCiRunner()
+        runner.prepareMetadata(uploadOptions)
+        
+        self.assertEqual(os.path.exists(f"{tempDir}/sigrid-metadata.yaml"), True)
+        with open(f"{tempDir}/sigrid-metadata.yaml") as f:
+            self.assertEqual(f.read(), "metadata:\n  externalID: \"1234\"\n")
+            
+    def testIgnoreEmptyEnvironmentVariables(self):
```
patveck, line 430: How about when in the Github action, a parameter is set to `null`? E.g., when someone wants to _delete_ the external ID, I'd expect that this `sigrid-metadata.yaml` is generated:
```yaml
metadata:
  externalID: null
```

Do we need a test for that?

dennis-sig: Eh… fair point, because I don't know if GitHub Actions even allows parameters to be `null`. How do we handle empty strings in the back-end? Are they considered the same as `null`, or will it actually set the external ID to an empty string?

dennis-sig: @patveck I have doubts if we should do this. In GitHub Actions, people do this:

```
divisionname: ${{ inputs.divisionname }}
```

Their notation is glorified Bash, so there is no real different between "value is null" and "value is empty" and "value is not set". I think it would be too error prone to let people un-set metadata in this way. I propose to leave the ability to un-set fields programmatically to the YAML file and end point, but to not support it herre.

patveck (approved):

PR #148, opened 2022-12-06
Add new team names field to metadata.

dennis-sig:


docs/sigrid-api-documentation.md
```diff
@@ -221,7 +223,7 @@ The response format of the customer-level endpoint (`GET https://sigrid-says.com
 ]
 ```
 
-All properties can be null except for `supplierNames` (which is always an array, but possibly empty) and `isDevelopmentOnly` (which is always true or falls).
+All properties can be null except for `supplierNames` and `teamNames` (which are always an array, but possibly empty), and `isDevelopmentOnly` (which is always true or falls).
```
bruno-oliveira, line 226: Small typo here: true or false

dennis-sig: Eagle eyes, fixed :)

PR #152, opened 2022-12-07
Fix broken link and clarify that we are using CycloneDX as SBOM standard

janlaan:


dennis-sig:
@janlaan GitHub Actions doesn't allow you to run the pipeline because you're not part of the SIG GitHub organization. The pipeline is fine, but we need to hit a button to make you able to run it.

PR #155, opened 2022-12-13
Experimental Markdown report.

dennis-sig:


dennis-sig:
This was an experimental branch, but we have since agreed to do this for real, which ended up being https://github.com/Software-Improvement-Group/sigridci/pull/267.

PR #157, opened 2022-12-21
Update checkout action in both documentation and usage.

dennis-sig:


MichielCuijpers (approved):
Lets work on informing our CI users that have an older version about this update

PR #160, opened 2022-12-23
Warn if someone literally uses the example name.

dennis-sig:


dennis-sig:
No longer relevant after Bruno added some extra characters to the example names that effectively bans people from copy/pasting them.

PR #165, opened 2023-01-13
Use temporary file for ZIP, which is removed afterwards.

dennis-sig:


sigridci/sigridci.py
```diff
@@ -140,17 +141,19 @@ def retry(self, operation, *, attempts=5, allow404=False, allowEmpty=True):
...
-        uploadLocation = self.obtainUploadLocation(systemExists)
-        uploadUrl = uploadLocation["uploadUrl"]
-        analysisId = uploadLocation["ciRunId"]
-        log(f"Sigrid CI analysis ID: {analysisId}")
-        log("Publishing upload" if self.publish else "Submitting upload")
-        self.uploadBinaryFile(uploadUrl, upload)
+        
+        with tempfile.NamedTemporaryFile() as tempFile:
+            uploadPacker = SystemUploadPacker(options)        
+            uploadPacker.prepareUpload(options.sourceDir, tempFile.name)
```
Geertien, line 147: I believe neither `tempfile.NamedTemporaryFile()` nor making a zip-archive out of the file in `uploadPacker.prepareUpload` adds the `.zip` extension to the file. So that makes the test checking for any files ending with `.zip` not very useful.

Although if I'm not mistaken the name of this (temporary) file is not important for the actual upload via the request, it wouldn't hurt either to have it end in `.zip` and that would clearly show the intention. The test would still be mainly testing the behaviour of `tempfile`, but as it still documents the desired behaviour/conscious decision I think that's fine. I believe the rest of the name we gave the original persisting `upload` file is not that important.

So I would suggest to add a suffix.

```suggestion
        with tempfile.NamedTemporaryFile(suffix=".zip") as tempFile:
            uploadPacker = SystemUploadPacker(options)        
            uploadPacker.prepareUpload(options.sourceDir, tempFile.name)
```

Geertien (changes requested):
Looks good to me in general. I did wonder if some users would like to keep te zip for their own administration and/or some other reason. I expect not though, so I think it is sufficient to build that option if it's really requested.

Then only the one comment remains.

dennis-sig:
@Geertien I accepted your suggested change, let's see if we can merge this on Monday morning.

Geertien (approved):
Great, looks good to me now 🙂

PR #166, opened 2023-01-18
Support the new subsystem option.

dennis-sig:


nicorikken:
@dennis-sig is this supposed to be used to do partial uploads of a system? That would be a really interesting addition for multi-repo systems where each repo can be uploaded as a subsystem. Or perhaps I misunderstood the intention of this code.

docs/faq.md
```diff
@@ -176,9 +180,14 @@ In some cases, the Sigrid CI client script might produce an error message `[SSL:
...
+If your repository is part of a larger system you can use subsystem to only get feedback on your repository. This is
+done on a best effort basis and you might get unexpected results if another subsystem published to Sigrid while your
+pipeline was running. To analyze subsystems Sigrid CI uses both the source code and measurement results from the last
+published measurement. If something is published while we retrieve these discrepancies might occur. If this happens, try
```
bruno-oliveira, line 188: Maybe we can re-write more clearly as:

"If results are published while we retrieve these measurements, some discrepancies might occur."

docs/client-script-usage.md
```diff
@@ -36,7 +36,7 @@ Notes:
...
+5. The `--subsystem` option can be used in cases where your repository is part of a group of repositories that are a single system in Sigrid. For example, you might have combined your back-end and front-end repositories within a single system in Sigrid, so that in Sigrid there are two top-level folders: `backend` and `frontend` containing the contents of your two repositories. However, you still want to get specific feedback for your front-end repository in Sigrid CI. In this case you would use `--subsystem frontend` so that Sigrid CI knows the location of your repository within the larger system.
```
bruno-oliveira, line 39: I propose: 
"The `--subsystem` option can be used when your repository is part of a group of repositories that is considered as being a single system in Sigrid.

For example, you might have combined your back-end and front-end repositories into a single system in Sigrid, such that in Sigrid there are two top-level folders: `backend` and `frontend` containing the contents of both repositories. However, you still want to get specific feedback for your `frontend` repository in Sigrid CI. In this case, you would use `--subsystem frontend` so that Sigrid CI knows the location of your repository within the larger system."

sbrinkhorst:
@nicorikken Yes, that is exactly the intention. The documentation is a bit minimal at this moment, we're probably going to extend that at a later moment. We wanted to take it for a spin at a few users first to see if there are any scenario's we haven't thought about.

nicorikken:
Will give it a try next week. Having looked to the code briefly there don't seem to be name requirements on the subsystem name as is the case for the system (length and allowed ascii characters).

PR #168, opened 2023-01-26
Document new scope options

dennis-sig:


docs/analysis-scope-configuration.md
```diff
@@ -123,14 +123,19 @@ The `thirdpartyfindings` section supports the following options:
     architecture:
       enabled: true
       
+Architecture Quality also requires the repository history to be included in the upload. This requires the `--include-history` option to be enabled in the [Sigrid CI client script](client-script-usage.md).
```
patveck, line 126: Would it be possible to link to an anchor inside `client-script-usage`?

patveck (approved):

docs/analysis-scope-configuration.md
dennis-sig: Theoretically yes, but we don't actually have a section for that yet :) We can do that as a follow-up, we can also provide some more information on exactly what information is sent when you enable that option.

PR #169, opened 2023-01-31
Revert "Use temporary file for ZIP, which is removed afterwards."

Geertien:
Reverts Software-Improvement-Group/sigridci#165


MichielCuijpers:
Sorry to revert, we noticed that this change was problematic on Windows. "Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later)."

PR #171, opened 2023-02-03
Create upload zip in temporary directory

Geertien:


test/test_sigridci_runner.py
```diff
@@ -520,7 +520,22 @@ def testFallbackToDefaultTargetIfNoSigridObjectives(self):
...
+        options = UploadOptions(sourceDir=tempDir)
+        target = TargetQuality("/tmp/nonexistent", 3.5)
+        apiClient = MockApiClient(publish=False)
+
+        runner = SigridCiRunner()
+        runner.run(apiClient, options, target, [])
+
+        for root, dirs, files in os.walk("."):
+            for file in files:
+                self.assertFalse(file.endswith(".zip"))
```
Geertien, line 537: As with the similar test in the previous attempt, this of course doesn't actually check the behaviour of the temporary directory (which we also shouldn't be testing in our code), but it does document and check the behaviour that we do not create a persisting zip file.

PR #179, opened 2023-03-22
Clarify who needs a token

dennis-sig:


docs/authentication-tokens.md
```diff
@@ -2,6 +2,11 @@
 
 Using Sigrid CI requires an *authentication token*. You can create and manage these tokens from your Sigrid account, using the instructions provided below.
 
+## Who should create the token?
+
+- **If you want to use Sigrid CI for an existing system that is already part of your Sigrid account:** Any Sigrid user account with access to that system can create the token. 
+- **If you want to on-board new systems to your Sigrid account:** On-boarding systems requires additional rights. This requires a Sigrid user account that is either an administrator, or a Sigrid user account that is authorized to access all systems. Sigrid user accounts with access to certain systems *cannot* on-board systems.
```
bruno-oliveira, line 8: Maybe in the last sentence we can use: `Sigrid user accounts with access *only* to certain systems *cannot* on-board systems.`

Emphasis on the `only` part to make it more clear that what can be seen as a "regular" account will not have rights to onboard systems.

dennis-sig: Phrased it slightly differently, how about this?

bruno-oliveira: Looks good to me!

PR #184, opened 2023-04-18
Hide —dummy argument from help.

dennis-sig:
@MichielCuijpers As discussed, this was taken from https://docs.python.org/3/library/argparse.html#help:

```
[argparse](https://docs.python.org/3/library/argparse.html#module-argparse) supports silencing the help entry for certain options, by setting the help value to argparse.SUPPRESS:
```


MichielCuijpers:
lgtm

PR #191, opened 2023-05-04
Fix link to api documentation

oscrx:
I found a broken link in the docs.


dennis-sig:
Thank you @oscrx ! We actually have a unit test that tries to prevent dead links, but this one was missed due to the anchor in the link. We'll update the test, but appreciated you took the time to make a pull request!

PR #193, opened 2023-05-08
Tweak CSS for SIG logo.

dennis-sig:


dennis-sig:
The location didn't change, but I made it absolute. Otherwise, the image path wouldn't work when you're in a subdirectory. We have tests to check against dead links in documentation, for the next batch we can make a test to make sure image paths are not relative.

dessens0 (approved):

PR #210, opened 2023-06-06
Moving portfolio overview part --> maintainability. Change filter icon.

syllosigm:


dennis-sig:
@syllosigm I suggest we move the `style="float:right;width="25 style=float:left"` to CSS. It will not be very fun to update every single image if we ever change the font size.

PR #215, opened 2023-06-08
Fix inline pictures, added menu visual.

syllosigm:
Fixed inline pictures, added menu screenshot/star rating legend.


dennis-sig:
Two comments:
- The system maintainability page contains empty sections.
- The quality models page should not mention the Performance and Reliability documents, since they only relate to SIG consultancy services and not to Sigrid.

dennis-sig:
Looks good to me! Do you have any further comments @bruno-oliveira ?

dennis-sig (approved):

PR #226, opened 2023-06-16
Added menu cross-references, started with security text.

syllosigm:


dennis-sig:
@syllosigm Looks good! At some point we need to determine where these FAQ pages should go, but for now we can leave them in the current structure. I will wait with merging this until after we have the new menu, but that will probably be no later than tomorrow.

PR #230, opened 2023-06-23
Revert "Use accordion menu."

dennis-sig:
Reverts Software-Improvement-Group/sigridci#227


dennis-sig:
Reverting because the new menu doesn't match some of the new items.

PR #240, opened 2023-06-30
Improving sso docs

MichielCuijpers:


MichielCuijpers:
@dennis-sig  improved SAML information

PR #278, opened 2023-08-09
Introduce separate entry point script, purely for reporting feedback.

dennis-sig:


sigridci/sigridci/junit_format_report.py
```diff
@@ -19,8 +19,11 @@
 
 class JUnitFormatReport(Report):
 
+    def __init__(self, outputDir="sigrid-ci-output"):
```
tukraus, line 22: I'm a bit on the fence here, should we move this up into the Report class? Now we specify the default value in 3 subclasses. I think the other subclasses that write to stdout would just ignore this outputDir field. Alternatively, perhaps just put "sigrid-ci-output" as a string constant in the Report superclass and use that in the file-writing subclasses.

test/test_sigridci_runner.py
```diff
@@ -43,62 +43,47 @@ def testForceLowerCaseForCustomerAndSystemName(self):
...
-
-        self.assertFalse(SigridCiRunner.isValidSystemName("noot", "aap_noot"))
-        self.assertFalse(SigridCiRunner.isValidSystemName("noot", "a"))
-        self.assertFalse(SigridCiRunner.isValidSystemName("noot", "$$$"))
-        self.assertFalse(SigridCiRunner.isValidSystemName("noot", "-aap"))
-        self.assertFalse(SigridCiRunner.isValidSystemName("noot", "a" * 65))
-        self.assertFalse(SigridCiRunner.isValidSystemName("noot", "aap--aap"))
-        
+
     def testValidateToken(self):
```
tukraus, line 50: Move this test case to dedicated test class of SigridApiClient?

tukraus (changes requested):

test/test_sigridci_runner.py
dennis-sig: OK. Most of the actual API interaction is covered by integration tests, but you're right this logic doesn't belong to the runner itself.

sigridci/sigridci/junit_format_report.py
dennis-sig: Let's go for the latter. The downside of making it an optional parameter in the super constructor, is that it would then *also* need to be an optional argument in all of the child classes. Having it as a field will avoid that repetition.

dennis-sig: Something I'm not sure about: Do we want to use `PipelineSummaryReport` in this new entry point? Because that will return an exit code, but depending on the way you process this, you might not want it to do `sys.exit(1)`.

tukraus (changes requested):

tukraus (approved):

PR #285, opened 2023-08-16
Catch error 502 that is thrown for analysis errors

sbrinkhorst:
Currently we return 502 without message, this should be improved but at least handle it correctly for now.

### DIFF
--- a/sigridci/sigridci/sigrid_api_client.py
+++ b/sigridci/sigridci/sigrid_api_client.py
@@ -77,6 +77,9 @@ class SigridApiClient:
                     else:
                         UploadLog.log(f"The system no longer exists (HTTP status {e.code})")
                     sys.exit(1)
+                elif e.code == 502:
+                    UploadLog.log("Something went wrong during analysis on our end (possibly related to configuration issues)")
+                    sys.exit(1)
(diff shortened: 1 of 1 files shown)


sigridci/sigridci/sigrid_api_client.py
```diff
@@ -77,6 +77,9 @@ def retry(self, operation, *, attempts=5, allow404=False, allowEmpty=True):
                     else:
                         UploadLog.log(f"The system no longer exists (HTTP status {e.code})")
                     sys.exit(1)
+                elif e.code == 502:
+                    UploadLog.log("Something went wrong during analysis on our end (possibly related to configuration issues)")
```
patveck, line 81:
```suggestion
                    UploadLog.log("Something went wrong during analysis on our end (could be an issue in `sigrid.yaml`, please review)")
```

patveck (changes requested):

dennis-sig:
@sbrinkhorst Maybe "at SIG" instead of "on our end"? In some cases it's not even known to developers that this is SIG's code running in their pipeline, as they are just using their internal pipeline template.

PR #293, opened 2023-08-31
Add the include flag to include folders instead of excluding

MarcW1g:
The current script contains the `--exclude` flag to exclude paths from the upload. This PR introduces a new flag called `--include` which instead only includes certain folders. This can be used to limit the upload to certain folders of the repository.

`--inlcude` and `--exclude` should not be used together.


dennis-sig:
Nice! Could you also document the new option in `docs/reference/client-script-usage.md`?

You will also need to add the new option to `action.yaml`, otherwise it doesn't work in GitHub Actions. Unfortunately they want you to manually list all options in the YAML file, it's not possible to do this dynamically.

dennis-sig (changes requested):
Other than the minor nitpicks this looks great!

MichielCuijpers:
well done @MarcW1g  this is good for the client that we talked to yesterday!

dennis-sig:
@MarcW1g Actually… you now restrict —include and —exclude so that they are not used together, but we could still allow that. The way you implemented it would mean "all files that are specified in —include except the files specified in —exclude". That's not bad, actually.

dennis-sig (approved):

PR #300, opened 2023-09-04
Update usage statistics for new Matomo version.

dennis-sig:


docs/_includes/stats.js
```diff
@@ -1,9 +1,13 @@
 document.addEventListener("DOMContentLoaded", () => {
```
bruno-oliveira, line 1: LGTM!

bruno-oliveira (approved):
This looks good to me, and we can merge it!

PR #304, opened 2023-09-14
Update authentication-tokens.md

MichielCuijpers:
@dennis-sig extending the info on when the notification emails for expired tokens are sent


dennis-sig (approved):

dennis-sig:
That was fast, grazie mille!

PR #307, opened 2023-09-21
Draft documentation for importing external security findings

janlaan:


dennis-sig:
@janlaan May I suggest we just create a separate section for this called "integrations"? That would get rid of the not-very-elegant subsubpages, and in general we tend to separate the analysis/integration stuff from the Sigrid website stuff in the navigation.

janlaan:
> @janlaan May I suggest we just create a separate section for this called "integrations"? That would get rid of the not-very-elegant subsubpages, and in general we tend to separate the analysis/integration stuff from the Sigrid website stuff in the navigation.
> 
> Also, it's not possible to run the pipeline from a fork, unfortunately this needs to be a branch within this repository.

@dennis-sig Makes sense. Like this?
I suppose this is where we can also put our other integrations, like our REST API (currently under "Reference"), and whatever we build in the future.

image

About running the pipeline, maybe @syllosigm can help putting this change in a branch in this repo, I don't have the access.

dennis-sig:
@janlaan Yes, exactly! Also, based on that statement the "publishing your dependencies" sub-page is in the wrong location, could you also move that to this new section?

If you and Sylvan are OK with the rest, I can make a branch out of your fork.

PR #309, opened 2023-09-27
Document external SAST integrations

dennis-sig:


dennis-sig:
@syllosigm @janlaan The tests did actually find a dead link. Could you correct it? Afterwards Sylvan will need to approve, because I created this one I can't approve it myself.

janlaan:
Do we also want to move the REST API page over to the integrations section?

PR #315, opened 2023-10-11
Update GitHub Marketplace docs.

dennis-sig:
@Geertien I also updated the documentation for the GitHub Marketplace version. Turns out it's just the exact same. See https://github.com/dennis-sig/exploration-day/pull/2/files for an example (you can check the YAML change since that is also in that pull request).

### DIFF
--- a/docs/sigridci-integration/github-actions.md
+++ b/docs/sigridci-integration/github-actions.md
@@ -132,6 +132,12 @@ jobs:
           publishonly: true
         env:
           SIGRID_CI_TOKEN: "${{ secrets.SIGRID_CI_TOKEN }}"
+      - name: "Sigrid pull request feedback"
+        uses: mshick/add-pr-comment@v2
+        if: always()
+        with:
+          message-id: sigrid
+          message-path: sigrid-ci-output/feedback.md
 ```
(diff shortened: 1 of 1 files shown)


docs/sigridci-integration/github-actions.md
```diff
@@ -132,6 +132,12 @@ jobs:
           publishonly: true
         env:
           SIGRID_CI_TOKEN: "${{ secrets.SIGRID_CI_TOKEN }}"
+      - name: "Sigrid pull request feedback"
```
Geertien, line 135: I believe this shouldn't be added to this file but the one below this one for the pull request feedback 😉

Geertien (changes requested):
I think it looks good if you move it to the correct one 🙂

docs/sigridci-integration/github-actions.md
dennis-sig: Whoops, you're completely right :)

Geertien: Looks all good now 😄

Geertien (approved):
Nice to see the GitHub Marketplace variant working and documented now.

PR #319, opened 2023-10-16
Tweak emoji and text used to explain objective status.

dennis-sig:


sigridci/sigridci/report.py
```diff
@@ -76,7 +76,9 @@ def getSummaryText(self, feedback, options):
             return f"✅  You wrote maintainable code and achieved your Sigrid objective of {target}"
         elif status == ObjectiveStatus.IMPROVED:
             return f"↗️  You improved your code's maintainability towards your Sigrid objective of {target}"
+        elif status == ObjectiveStatus.UNCHANGED:
+            return "⏸️️  You didn't improve your code, but it also didn't get any worse"
```
patveck, line 80: The phrase "but it also didn't get any worse" suggests that the code isn't good to begin with, so it sounds very negative to me. For instance, you have a target of 3.5, you're currently at 3.4, and the current Sigrid CI analysis determines the rating did not change. Then I think the user doesn't deserve the "... any worse" bit. Here's a more neutral alternative:
```suggestion
            return "⏸️️  You didn't improve your code, and consequently the rating did not change"
```

sigridci/sigridci/report.py
```diff
@@ -76,7 +76,9 @@ def getSummaryText(self, feedback, options):
...
         elif status == ObjectiveStatus.WORSENED:
-            return f"❌  Your code did not manage to improve towards your Sigrid objective of {target}"
+            return f"⚠️  Your code did not manage to improve towards your Sigrid objective of {target}"
```
patveck, line 82: I don't think the "manage to" bit is needed, and in my ears it sounds overly negative. This would be the more neutral alternative:
```suggestion
            return f"⚠️  Your code did not improve towards your Sigrid objective of {target}"
```

test/test_markdown_report.py
```diff
@@ -205,7 +205,7 @@ def testNegativeSummaryWhenNoImprovement(self):
 
         report = MarkdownReport()
         summary = report.renderSummary(feedback, self.options)
-        expected = "**❌  Your code did not manage to improve towards your Sigrid objective of 3.5 stars**"
+        expected = "**⚠️  Your code did not manage to improve towards your Sigrid objective of 3.5 stars**"
```
patveck, line 208:
```suggestion
        expected = "**⚠️  Your code did not improve towards your Sigrid objective of 3.5 stars**"
```

test/test_pipeline_summary_report.py
```diff
@@ -42,7 +42,7 @@ def testDisplayLandingPageFromClient(self):
         report.generate("1234", feedback, self.options)
         
         expected = """
-            ** ❌  Your code did not manage to improve towards your Sigrid objective of 3.5 stars **
+            ** ⚠️  Your code did not manage to improve towards your Sigrid objective of 3.5 stars **
```
patveck, line 45:
```suggestion
            ** ⚠️  Your code did not improve towards your Sigrid objective of 3.5 stars **
```

patveck (changes requested):

sigridci/sigridci/report.py
dennis-sig: Fair point. There is another thing with this message: This is the only one that doesn't mention what your objective actually is. That's purely for aesthetics, it doesn't look nice when the title word-wraps. But ideally we would find a way to have a similar meaning as this suggestion, but in a way that *also* allows us to put the objective in there.

dennis-sig:
Not directly in response to a comment, but we revised the emoji after discussion with Janelle and Eva: The red cross was seen as a bit too aggressive, and the black "=" doesn't work well for people using dark mode.

PR #321, opened 2023-10-16
Add documentation of new portfolio-level query parameters

patveck:


patveck:
@dennis-sig My PR only changes documentation, but nevertheless a Python unit test fails. Can you have a look?

dennis-sig:
@patveck Could you merge master back into your branch? I think that should address both the conflict and the failing test.

dennis-sig (approved):

PR #326, opened 2023-10-23
Anonymize repository history before publishing it.

dennis-sig:


sigridci/sigridci/repository_history_exporter.py
```diff
@@ -27,16 +31,35 @@ def exportHistory(self, sourceDir):
...
+    def createHistoryExportFile(self, history, outputFile):
+        with open(outputFile, "w") as f:
+            for line in history.strip().split("\n"):
+                f.write(self.anonymizeHistoryEntry(line))
+
+    def anonymizeHistoryEntry(self, gitLog):
+        anonymized = ""
+
+        for line in gitLog.strip().split("\n"):
+            if line.startswith(("@@@", "'@@@")):
```
patveck, line 57: Is there a way to make the code inside this block more self-explanatory? In its current form, it is clear it is about anonymization, but not about _what_ we're anonymizing. The only thing I could think of would be to introduce "unnecessary" variables, something like:
```python
author = cell[2]
committer = cell[3]
```
But that's just me guessing those cells are actually the author and committer, respectively.

sigridci/sigridci/repository_history_exporter.py
```diff
@@ -12,13 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import hashlib
```
patveck, line 15: I think using hashing is perfectly OK. Normally, hashing is insufficient: someone who knows the set of all possible values being hashed (e.g., a list of employees) could just hash that list and compare the results with our anonymized values. However, nowhere in Sigrid do we display the resulting hashes, they're only used internally for grouping. That way neither us nor our users can circumvent this process.

sigridci/sigridci/repository_history_exporter.py
dennis-sig: Python has a somewhat obscure syntax for this. Would you be OK with the following:

```
aap = ["a", "b", "c"]
a, b, c = aap.split(",")
```

sigridci/sigridci.py
```diff
@@ -69,8 +69,10 @@ def parseTarget(target):
     parser.add_argument("--exclude", type=str, default="", help="Comma-separated list of files/directories to exclude.")
     parser.add_argument("--include", type=str, default="", help="Comma-separated list of files/directories to include.")
     parser.add_argument("--showupload", action="store_true", help="Logs the contents of the upload published to Sigrid.")
-    parser.add_argument("--include-history", action="store_true", help="Publish repository history to Sigrid.")
     parser.add_argument("--sigridurl", type=str, default="https://sigrid-says.com", help=SUPPRESS)
+    # This option is now enabled by default, using an anonymized history,
+    # but we leave it here to avoid breaking people's configuration.
```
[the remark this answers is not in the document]
dennis-sig: That is actually not entirely true. The reason for leaving the argument is actually so that people don't get a "unknown argument —include-history" error.

sigridci/sigridci/repository_history_exporter.py
patveck: Obscure it is, but it makes it much clearer what we're doing!

patveck (approved):

sigridci/sigridci.py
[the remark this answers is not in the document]
patveck: But wouldn't it be a good idea to have an escape hatch, just in case this needs to run, if even temporarily, in an environment where `git` is unavailable and/or cannot be used? Of course, I can think of a hundred thousand more things for which one day an escape hatch might be handy, but this one we already have, so we can as well keep it.

PR #330, opened 2023-10-31
Handle dead symlink in upload.

dennis-sig:


dennis-sig:
Fixed by https://github.com/Software-Improvement-Group/sigridci/pull/439

PR #338, opened 2023-11-10
Documentation search

dennis-sig:


docs/_includes/search.js
```diff
@@ -0,0 +1,44 @@
+{%- assign index = "" | split: "" -%}
+
+{%- for page in site.html_pages -%}
```
bruno-oliveira, line 3: Maybe we could extract this to a small, descriptive function such as: `indexGeneration(site)`?

bruno-oliveira (approved):

docs/_includes/search.js
dennis-sig: I had a look, but based on http://hamishwillee.github.io/2014/11/13/jekyll-includes-are-functions/ you actually can't do that in the subset of Jekyll used by GitHub Pages. According to that page use includes as "hacky" functions, which is also whatI'm doing here.

bruno-oliveira: Ah then we will know for the future!

PR #344, opened 2023-11-29
Extend documentation for SFTP, manual uploads, Snapshot Converter.

dennis-sig:
@valdore86 This adds both the Snapshot Converter caveat to the technology support list, and also adds the "classic" upload documentation. Since you had a hand in both discussions, would you be willing to review it?


valdore86 (approved):
All good! Thank you for bundling the original files into a comprehensive instruction covering it all 😊

PR #345, opened 2023-11-29
Be more tolerant in exporting commit message.

dennis-sig:


dennis-sig:
@patveck A while ago we added the multi-assign to make the git log parsing more readable. That fails if you have more fields in a line, hence this adding the `*rest` thing. This does mean we don't get the full commit message in those cases, but we can live with that since we don't use it for the analysis anyway.

PR #361, opened 2024-01-12
Support convert option.

dennis-sig:


sigridci/sigridci/sigrid_api_client.py
```diff
@@ -117,8 +117,12 @@ def obtainUploadLocation(self, systemExists):
         elif self.options.runMode == RunMode.FEEDBACK_AND_PUBLISH:
             path += "/publish"
 
-        if self.subsystem:
+        if self.subsystem and self.options.convert:
```
sbrinkhorst, line 120: In the init (on line 40) there is`self.subsystem = options.subsystem`. The same is not done for the convert option. That implies that the assignment in the init is not needed and we could also use `self.options.subsystem` here. Or at least we should make it consistent because now it makes you wonder how these things are different.

dennis-sig: No you're right, there was no good reason for that.

sbrinkhorst (approved):
Looks good!

PR #380, opened 2024-02-12
Fix --include option inadvertently disabling history.

dennis-sig:
Using the semi-new  `—include` option by @MarcW1g inadvertently disabled the Git history for Architecture Quality, unless you explicitly add git.log to your includes. Note `endswith` accepts a tuple, hence the `(…)` (also in case we want to extend the list in the future).

### DIFF
--- a/docs/reference/release-notes.md
+++ b/docs/reference/release-notes.md
@@ -3,9 +3,11 @@ Sigrid release notes
 
 SIG uses [continuous delivery](https://en.wikipedia.org/wiki/Continuous_delivery), meaning that every change to Sigrid or the underlying analysis is released once our development pipeline has completed. On average, we release somewhere between 10 and 20 times per day. This page therefore doesn't list every single change, since that would quickly lead to an excessively long list of small changes. Instead, this page lists Sigrid and analysis changes that we consider noteworthy for the typical Sigrid user.
 
-### February 5, 2024
+### February 12, 2024
 
 - **Sigrid CI:** The output directory where Sigrid CI saves feedback is now a configurable. See the [options reference](client-script-usage.md) for details.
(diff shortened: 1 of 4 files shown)


docs/reference/release-notes.md
```diff
@@ -3,9 +3,11 @@ Sigrid release notes
 
 SIG uses [continuous delivery](https://en.wikipedia.org/wiki/Continuous_delivery), meaning that every change to Sigrid or the underlying analysis is released once our development pipeline has completed. On average, we release somewhere between 10 and 20 times per day. This page therefore doesn't list every single change, since that would quickly lead to an excessively long list of small changes. Instead, this page lists Sigrid and analysis changes that we consider noteworthy for the typical Sigrid user.
 
-### February 5, 2024
+### February 12, 2024
 
 - **Sigrid CI:** The output directory where Sigrid CI saves feedback is now a configurable. See the [options reference](client-script-usage.md) for details.
+- **Open Source Health:** The Sigrid user interface now shows whether dependencies are direct or transitive. This is helpful when using the [transitive option for Open Source Health](analysis-scope-configuration.md#open-source-health), as it allows you to quickly determine a dependency's origin.
```
patveck, line 9: I think we currently only support this for NPM and Maven. Whenever we add support for other ecosystems, we can add a new entry to the release notes.
```suggestion
- **Open Source Health:** The Sigrid user interface now shows whether dependencies are direct or transitive (currently only for Maven and NPM). This is helpful when using the [transitive option for Open Source Health](analysis-scope-configuration.md#open-source-health), as it allows you to quickly determine a dependency's origin.
```

patveck (changes requested):

patveck (approved):

PR #386, opened 2024-02-15
Log last 4 characters of token.

dennis-sig:


erik-blokland (approved):
Extra log line and test changes look fine to me

PR #387, opened 2024-02-16
Include instructions to upload git history in architecture FAQ

janlaan:


janlaan:
@dennis-sig

dennis-sig:
@janlaan I find the description a bit confusing. It implies you need to actively do something to get the Git history into Sigrid, which is not true for like 95% of people using Sigrid CI. I would suggest to change the list:

1. If you're using Sigrid CI using the examples provided in this documentation, your Git history is automatically anonymized and published to Sigrid.
2. If you are *not* using Sigrid CI.
  - (your points)

dennis-sig (changes requested):
See my suggestion, that will help to avoid confusion.

dennis-sig:
Ended up being turned into @MichielCuijpers 's proposal.

PR #390, opened 2024-02-19
Run unit tests for pull requests created forks.

dennis-sig:
- We *do* want to run the Python unit tests for pull requests in forks.
- We *cannot* run Sigrid CI for forks, since we don't want to give them access to Sigrid. Note we run Sigrid CI twice, we test with both the Docker version and the "plain" version.
  - Those two Sigrid CI runs were actually located in different actions, which was a bit weird, so I moved them together.


tukraus (approved):
👍

PR #408, opened 2024-03-19
Methodology osh

lbergmans:
Create documentation for guiding organizations in how to handle OSH. First document in the methodology series.


docs/workflows/best-practices-osh.md
```diff
@@ -0,0 +1,398 @@
...
+
+When a vulnerability is found, it will be remediated within a specified time period.
+  The table below is a suggestion how to specify this, depending on the criticality of the system:
+  | CVSSv3 Range | Label | Remediation Deadline Critical | Remediation Deadline Regular |
+  | --- | --- | --- | --- |
+  | 9.0 – 10.0 | Critical | Within 1 working day | Within 1 working day | 
+  | 7.0 – 8.9 | High | Within 14 days | Within 30 days |
+  | 4.0 – 6.9 | Medium | Within 60 days | Within 90 days |
+  | 0.1 – 3.9 | Low | Within 1 year | Within 1 year |
+  > TODO: check that the above table makes sense, especially the added last column
```
dessens0, line 238: I think you only need this table for public-facing systems, with just one column (like we use it ourselves)

docs/workflows/best-practices-osh.md
```diff
@@ -0,0 +1,398 @@
...
+  - [In](http://3.in/) extreme cases, we will shut down the application until the vulnerability is resolved.
+  - note: when addressing the vulnerabilities of a specific library by updating to a newer version, that always also improves the freshness
+  
+
+### 6. Handling detected license issues [ROUGH DRAFT]
+- Libraries must have an acceptable license.
+  - This may be a paid license or an acceptable open-source license. [Company] maintains a list of common licenses used in free and open-source software (FOSS); if a library has a license listed as acceptable, it can be used. Otherwise, see if an alternative is available, or contact the [applicable role] to discuss whether the license is acceptable.
+- See also the License Risk Evaluation Framework slide at the bottom of the wiki page: [OSH interpret results wiki page: https://softwareimprovementgroup.atlassian.net/wiki/spaces/SSM/pages/50317951024/How+to+interpret+Open-Source+health+findings](https://workflowy.com/#/c9c4745b6b6b)
```
dessens0, line 251: In our documentation, we cannot refer to the wiki (at least, we need to make it clear that 'regular' users can't use the link.)

docs/workflows/best-practices-osh.md
```diff
@@ -0,0 +1,398 @@
...
+- Note that the risk depends on the context: 
+  - e.g. when developing open-source software, more licenses are acceptable.
+  - it also depends on how you use a library: it is sometimes needed/a solution to wrap the library in an executable component that can be used by calling, instead of becoming a part of the codebase.
+
+
+### 7. Handling detected lack of freshness [ROUGH DRAFT]
+- The teams are responsible for keeping libraries reasonably up to date: 
+  - small library updates can be updated as part of regular maintenance; 
+  - larger updates (e.g. major versions or frameworks) should be planned explicitly as part of a development sprint.
```
dessens0, line 260: I would drop the 'as part of a development sprint'. You still need to plan if you don't use sprints.

docs/workflows/best-practices-osh.md
```diff
@@ -0,0 +1,398 @@
...
+### 8. _Handling detected lack of activity_ (TBD)
+
+## Handling your libraries
+
+
+### 9. Updating a library [ROUGH DRAFT]
+There can be several reasons to consider updating a library.
+
+Checklist before updating:
+- Does it have open issues?​
```
dessens0, line 274: Not sure what this check is meant to cover. A well-maintained library [will have open issues.](https://github.com/spring-projects/spring-boot/issues). Zero open issues typically equates to low activity. I would say 'Check the release notes to see if and how you might be affected by the changes.'.

docs/workflows/best-practices-osh.md
```diff
@@ -0,0 +1,398 @@
...
+- Is the license (still) compatible?​
+  - → [Handling detected license issues](#handling-detected-license-issues)
+- Does it contain optimizations for performance, maintainability, functionality?​
+  - ? not sure why this is exactly asked?
+- Is the new version compatible?  ​
```
dessens0, line 280: I would drop these checks. This is covered by checking the notes.

docs/workflows/best-practices-osh.md
```diff
@@ -0,0 +1,398 @@
...
+- How mature is the version?​
+- Is it stable enough?​
```
dessens0, line 282: I would drop these. I wouldn't know how to meaningfully do that. I would say something like 'Use a stable version unless there is a real reason not to do so. An example might be that a Release Candidate fixes a vulnerability and you don't want to wait for the stable version to come out'.

docs/workflows/best-practices-osh.md
```diff
@@ -0,0 +1,398 @@
...
+
+Always Manage libraries with a package manager (don’t mix third party code with first party code)​
```
dessens0, line 284: This sentence feels a bit lost here. I'd be inclined to cover the use of a package manager once and keep it simple. 'If your ecosystem allows it, use a package manager'.

docs/workflows/best-practices-osh.md
```diff
@@ -0,0 +1,398 @@
...
+
+- compatibility break
```
dessens0, line 286: I'm a bit unclear about what is meant by 'technologies' here. 
- You were using a (deprecated) method that's no longer available. Your IDE will likely make you aware of this.
- A nastier version is that you weren't using the library correctly and the new version is less forgiving but without alams going off.
- The library needs transitive dependencies that have become incompatible with needs of other libraries

docs/workflows/best-practices-osh.md
```diff
@@ -0,0 +1,398 @@
...
+  - Compatibility breaks can be caused by the framework/library no longer supporting the technologies used in the system.​
+  - Analyze the impact and consider the following things:​
+  - What is the impact of updating the system to make it compatible?​
+  - Are there alternatives of the library?​
+  - Are the technologies used still relevant for the future of the system?​
+
+- when is postponing updates a good idea?
+  - The longer you postpone updating, the bigger the eventual pain. As your system grows and evolves, the costs and risks of upgrading an old library increase. Such an accumulation of maintenance debt may lead to a much larger effort than in the case of smaller incremental updates.​
+  - For core functionalities, it is a good idea to stay on a stable version of the library. ​Being an early adopter of a new version comes with some risks. Bugs are common in immature versions of libraries as they are not thoroughly tested. Although you have the opportunity to contribute, by testing and potentially providing solutions to bugs you encounter in open source libraries, it might be a better practice to wait until the library becomes more mature.​​
```
dessens0, line 296: My experience is a bit different. At least in the Java ecosystem, popular libraries are fairly mature so if a new, stable version comes out it typically works. The typical workflow is that once a new major version comes out, you start testing it to assess the impact. Meanwhile, other people do the same and any serious problem that surfaces receives a patch-level fix. So after a couple of weeks, you know where you stand. This is typically one of
- Go ahead and update
- There's a compatibility break which requires fixing on your end, so plan the work
- There's an acknowledged compatibility break which requires fixing on their end. Wait and keep an eye on the releases.

Bottom line: if a new, stable version comes out: don't wait, start testing. If it's really core and really important, already start testing with release candidates.

docs/workflows/best-practices-osh.md
```diff
@@ -0,0 +1,398 @@
...
+
+- First, check whether a newer version of the library may solve the issue, then consider updating ([Updating a library]); you may also wait a bit until a fix is being released, especially when the issue is being worked on.
+- If the issue is a bug, you can look at the code of the library and develop a bug fix: 
+  - You may be able to wrap the relevant library call(s) with extra code that corrects or hides the bug.
+  - Alternatively, you can temporarily use the modified library while merging back the bug fix into the library (be aware of the license). Once the community has accepted the fix, you can update and remove the local code​.
+
+- If no other solution is feasible and a modification is absolutely required (or: the costs of alternative solutions are very high), the source code should be 'adopted' (if permitted by its license) and put into a designated area in a version control system. The modification should be documented so that it can be re-applied whenever a newer version of the library is made available.
+
+### 12. Replacing a library
```
dessens0, line 332: I feel there's a bit of redundancy in the structure. To me, it appears that we have the following:
- Sometimes, you need to switch libraries. This can be caused by insufficient health (license, activity, management), compatibility issues or functional requirements.
- Then, this is how you go about it (Replacing a new library). When it's not a replacement but a new library, do the same checks minus a couple that only apply to a replacement. (where is the old one used and how is the interface different?)

docs/workflows/best-practices-osh.md
```diff
@@ -0,0 +1,398 @@
...
+1. _Define the usage of a package manager_: choose the package manager(s) to be used, at least per system, preferably shared across the organization. Depending on the technologies that are used, you may need multiple package managers.
+   - The package managers need to be integrated in your CI/CD pipeline.
+  <!-- more useful info here: https://gurukuldevops.com/package-management-in-devops-tools-and-best-practices/ -->   
+
+2. _Set the thresholds for library risks_ that are (not) acceptable: this is applicable to all types of risks. Set these goals in the [Sigrid objectives](#TBD). Define goals for:
+   - The amount of and allowed types of vulnerabilities.
+   - The allowed licenses.
+   - Library freshness; the time since a new version of a library has been released.
+
+[Note Asma] Amount of vulnerabilites is not always a fitting metric. Having a goal to reach 0 critical and or high risk findings makes sense. Low and mediums are often present in larger numbers
```
dessens0: For security yes, but in my experience this is not the case for OSH. Most of the vulnerable libs we see are either 'high' or 'medium', with some 'low's sprinkled in. Reporting OSH vulnerabilities is relatively expensive with CVEs being issued and everything, so I guess that's not worth it for 'low' vulnerabilities.

dessens0: As a result, I rarely find myself actually using the vulnerability rating when making assessments. It's basically, is there a fix -> apply it. If not, understand why and take it from there. Most maintainers try to address open CVEs because people start asking questions about it.

docs/workflows/best-practices-osh.md
```diff
@@ -0,0 +1,398 @@
...
+  - For core functionalities, it is a good idea to stay on a stable version of the library. ​Being an early adopter of a new version comes with some risks. Bugs are common in immature versions of libraries as they are not thoroughly tested. Although you have the opportunity to contribute, by testing and potentially providing solutions to bugs you encounter in open source libraries, it might be a better practice to wait until the library becomes more mature.​​
+  - Don’t go for the "If it ain’t broke, don’t fix it" strategy​
+    - This strategy implies that you don’t update unless you have to. You stay with the current version of the third-party library until you notice something wrong in your application, no matter how often the vendor publishes an update. ​
+    - Whilst easier in the short term, with this strategy you will end up with a system which depends on outdated and unmaintained libraries, where you can't use some other libraries since they require a newer version of that library which you can't upgrade and at some point. You may lose the ability to fix some issues at all.​
+
+
+### 10. Adopting a new library [ROUGH DRAFT]
```
dessens0, line 302: I'd use the word 'use' instead of 'adopt'. We use 'adopting' elsewhere to indicate taking responsibility for maintaining the code, which is not what we try to say here, right?

docs/workflows/best-practices-osh.md
```diff
@@ -0,0 +1,398 @@
...
+5. _Declare which libraries should not be checked_
+   - This is useful when a library has properties that cause Sigrid to signal a risk, but that risk is a false positive. 
+   - Create an _ignore-list_. This list requires CISO approval. The ignore-list must be reviewed regularly (a few times per year): define when this review will happen.
+   - Do note that if a library is put on the ignore-list since the reported vulnerability-list is a false positive, that does not necessarily mean that the other types of risk should be ignored as well.
+   > NOTE: this was called the 'allow-list', but 'ignore-list' seems to carry the purpose better.
+
+   > TODO: (how) can customers set up an ignore-list for Sigrid?
+   > TODO: planned for Sigrid(?): instead of an ignore-list, be able to snooze/hide false positives from OSH → check
+
+6. Optionally: _Define a shared permitted-list_: it can be useful (and in some organizations required) to have a shared list of libraries that are permitted.
```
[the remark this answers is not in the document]
lbergmans: That should entail at least the elements from (current) 'adopt a library', plus perhaps some organisation specific rules (be starred minimum times, should not originate from some countries??, have passed your own SAST checks, ..). We could ask around, even.

docs/workflows/best-practices-osh.md
lbergmans: so, it does not matter otherwise?? I think security best-practice is that also non-public facing systems need to be secure (multiple lines of defense).

docs/workflows/best-practices-osh.md
lbergmans: indeed, this was a 'note to self'

docs/workflows/best-practices-osh.md
lbergmans: indeed, that is a copy-paste from somewhere.
Using a package manager is mentioned in 2. Define OSH related policies for development, the only question is, should we mention it here once more, just to emphasize this?

docs/workflows/best-practices-osh.md
lbergmans: I do not remember where this came from (I do hope I did not write this myself..); will rewrite using your suggestion. 
Are there known cases of language/SDK updates that are not backward compatible perhaps??

lbergmans: perhaps if your application is using older versions of the language, and a library is using a new (incompatible) one?

docs/workflows/best-practices-osh.md
lbergmans: makes sense: also here, I do not remember the source of this (plausable deniability ;-) )

docs/workflows/best-practices-osh.md
lbergmans: correct, I think this is better now

dennis-sig (approved):

PR #415, opened 2024-04-09
Avoid running Sigrid CI if the system has been deactivated in Sigrid.

dennis-sig:


MichielCuijpers:
yes thanks

MichielCuijpers:
Changing the way we handle deactivated systems will save SCC a lot of noise. 100+ tickets in  the last week 
Screenshot 2024-04-10 at 09 16 23

tukraus (approved):
👍

PR #421, opened 2024-04-15
Include feedback links in report.

dennis-sig:


dennis-sig:
@patveck The nice thing about Sigrid CI is that you can see the changes in this pull request. This adds the feedback links at the bottom.

docs/landing/feedback.html
```diff
@@ -52,15 +52,16 @@
         document.addEventListener("DOMContentLoaded", () => {
             const statsURL = "https://sigrid-says.com/usage/matomo.php";
             const params = new URLSearchParams(window.location.search);
+            const timestamp = new Date().getTime();
```
patveck, line 55: Does this result in a URL-friendly timestamp, preferably in ISO 8601 format? And should it be normalized to UTC (this might run anywhere on our planet in a CI/CD environment).

sigridci/sigridci/markdown_report.py
```diff
@@ -51,6 +51,12 @@ def renderMarkdown(self, analysisId, feedback, options):
         md += "\n----\n\n"
         md += f"- [**View this system in Sigrid**]({sigridLink})\n"
         md += f"- [**View this Sigrid CI feedback in Sigrid**]({landingPage})\n"
+        if options.feedbackURL:
+            md += "\n----\n\n"
+            md += "## Did you find this feedback helpful?\n\n"
```
patveck, line 56: It is clear from the source code here on Github that this is anonymous. Nevertheless, it might be wise to say so, something like: "Did you find this feedback helpful? (will be submitted anonymously)".

docs/landing/feedback.html
dennis-sig: It returns the timestamp as an integer. We don't actually use the timestamp for anything, this `t=` is a convention to "break through" proxies that might otherwise cache GET requests to images, even when the request sets the correct response headers to avoid caching.

sigridci/sigridci/markdown_report.py
dennis-sig: Well, it's not completely anonymous. It's true we (intentionally) don't know the name of the person, but we do record the system for which feedback is provided.

patveck (approved):

PR #428, opened 2024-04-24
OSH workflow feedback

janlaan:


lbergmans:
With so many changes, clearly I would disagree or dislike a bunch, but it seems not: thanks for all the improvements!

Ok, I managed to come up with 1 and a half (the latter a simple typo) remark/question: 
- In section 3 you added: "3. Create an abstraction layer between the dependency and your code, this isolates changes coming from a library update, and also makes it easier to replace the library completely.": do you advise to always do this? I wrote something to the same extent, but in section 12. Replacing a library, and also in section 11. When a library does not meet requirements; so my idea was: do this only when needed.
- typo: "libary" (section 5)

lbergmans (approved):
no more wishes from my side, let's go!

PR #429, opened 2024-04-25
added a footnote 

lbergmans:
... that OSH benchmarked star ratings are not yet in Sigrid, but coming soon.

regenerated locally to test that rendering footnotes works well:

image


dessens0 (approved):
That's an easy one.

PR #432, opened 2024-04-29
Update documentation layout

dennis-sig:


lbergmans (approved):
This is 2 steps forward; 
ideally we will have more than 1 aside job in the future; so one can distinguish 'tip', 'warning', 'example', etc.

PR #437, opened 2024-05-01
First version of organization guidelines

dessens0:


docs/workflows/best-practices-organization.md
lbergmans, line 1: ### Introduction
* "...how organizations should set themselves up to achieve high-quality software. We are aware that many models, ideas, frameworks and the like are available on this topic..": really? what are examples? I assume there is a lot on 'achieving quality' in general, but on how to organise things specific for high-quality software?
* "Rather, we identify three different *levels* of decision making.."; I think level is at this point not so clear (resolved in next section); I was thinking to use 'stakeholders' perhaps? 

### Three levels of decision making
* I would prefer another word instead of 'level' BTW
* does it make sense to move the 'how to identify' section to the start? this explains in more details what/who these levels are
* "..(or that aren’t effectively/efficiently addressed by the team)..."; not clear what you mean, can you give a few examples?
* "Engineering Leader Level": perhaps _start_ with use some example roles to make more clear who these people are? (tech leads, architects, QA people, ...?) now only at the end of the paragraph
 -> now only a* "...but are not directly related to feature work..."; perhaps drop this? not clear to me what the role of this text is?
* "...triggers for this _loop_..": no loop has been mentioned ?
* "They are the core players, supported by the team members and higher-level managers.​"; who are 'they', do you mean the people in the engineering leader level? core for what?
* "...but I'm not sure we often see that in practice"; reformulate?
* "Since there is no entrenched practice, there's not something we can easily tap into.": not clear what this means (also: does 'we' refer to SIG?)
* "... the concerns can also be represented by ..."; what does that mean?? 
 
### What is needed from each level
* I think the picture is great, esp. the concrete items listed at each level/stakeholder. I was a bit triggered that engineering leads lists _technical_ impediments only; there are many things related to how the work is organised, within teams--> that should be added to the teams (they are autonomous, so they should address many issues themselves), and between teams (that is where esp. engineering level would play a role?).
* Should enabling teams/platform teams also get into the picture somewhere? (oh it does, a bit hidden in "A hands-on tech lead"..)

### Set a target percentage
* "articulate the work the want to do ": 'the want' -> 'they want'

### Applying the boy scout rule
* "Bigger, more structural improvements need to be planned explicitly. If not, small changes may take a disproportional amount of time ": to avoid confusion, add "small _functional_ changes"?

### Embedding technical improvements in existing frameworks
* I was first assuming software frameworks...  -> 'process frameworks'??

NB: can we add renovate and rebuild scenarios? (not only at a system level, but esp. on a component level?)
Also: sometimes quality requirements are demanded top-down: by customers, or regulators (esp. security-related), in which case they may 'flow' more naturally via product managers etc.

docs/workflows/best-practices-organization.md
```diff
@@ -0,0 +1,97 @@
+# Guidelines on how to organize for quality software
```
dennis-sig, line 1: Looking at the file name, but you could call this best practices instead of guidelines?

docs/workflows/best-practices-organization.md
```diff
@@ -0,0 +1,97 @@
...
+
+**Management** can be harder because many organizations, especially larger ones, have multiple layers of management and not all of them are relevant in our context. Typically though, management represents the (strategic) organizational goals and has a budget responsibility. They are accountable for the outcomes of software development and as such, are concerned with topics like cost, speed and risk. While this accountability ultimately lies with a CEO, the concerns can also be represented by a product owner or project manager.
+
+**Engineering leads** are typically the hardest one to identify because it's of the three the least established and exists in various forms and to varying degrees. Engineering leads are typically involved in (higher-level) solution design, technology choices or security analyses and policies. Job titles may include tech lead, team lead or architect. Engineering leads may exist within teams and organize virtually through Communities of Practice, Guilds or Chapters. In other cases, separate engineering lead roles or teams exist, such as enterprise architects, security specialists or Centers of Excellence.
```
dennis-sig, line 32: I would make a comment that we refer to these people as the architect persona elsewhere in the documentation (to avoid every page using slightly different termininology).

docs/workflows/best-practices-organization.md
```diff
@@ -0,0 +1,97 @@
...
+
+This picture shows the interactions and responsibilities of the different groups:
+
+<img src="../images/organization-groups-interaction.png" width="700" />
+
+## Strategies to ensure technical improvements
+
+For the different groups to understand their responsibilities in producing quality software is one thing, to ensure that technical improvements are indeed happening is another. In this section, we present some strategies that make technical improvements part of the regular workflow. These strategies are generically applicable in the sense that they can be applied in an existing development process, whatever that process might be. We have seen these strategies work for customers and/or used them ourselves. Also, you can mix-and-match strategies and switch strategies over time as you learn more or circumstances change.
+
+In [another section](#embedding-technical-improvements-in-existing-frameworks), we discuss how to embed technical improvements in existing frameworks such as SAFe or SCRUM.
```
dennis-sig, line 59: Dead link?

docs/workflows/best-practices-organization.md
```diff
@@ -0,0 +1,97 @@
...
+
+The main disadvantage of this method is that it may create a 'moral hazard' in regular sprints, teams might focus exclusively on features (or feel forced to do so) under the assumption that any technical debt they may introduce is taken care of during the quality sprint.
+
+A way to mitigate this is to not have a regular '1 out of X' cadence for quality sprints, but have the team decide on the need to have one based on the technical improvement backlog. This obviously requires a fair amount of team discipline and maturity.
+
+## Embedding technical improvements in existing process frameworks
+Organizations may have implemented a framework to organize their software development. While the [strategies presented](#strategies-to-ensure-technical-improvements) are universally applicable, it may help to explain how to organize for quality software within these process frameworks, using its concepts and jargon.
+
+- [Agile/Scrum](../workflows/agile-development-process.md)
+- Scaled Agile Framework (SAFe)
```
dennis-sig, line 96: Not part of this page, but we do need a page for SAFe as part of our documentation at some point.

docs/workflows/best-practices-organization.md
dessens0: Will do

docs/workflows/best-practices-organization.md
dessens0: Yes, that's probably next on the list. I can perhaps leave out the text entirely until we have it.

docs/images/organization-groups-interaction.png
lbergmans, line 1: maybe another way to formulate the management role is that they need to set some of the contraints/boundaries for the development teams to operate within (such as security and licensing requirements that are must-haves, but probably also other standards that are needed to ensure interoperability and integration between teams, e.g. for issue tracking and code repos). The latter is about reducing impediments.

lbergmans (approved):
I think this is become a nice 'manual'; it makes sense to discuss this to quality teams such as at Alliander and belastingdienst; they may use the strategies towards the teams and/or have feedback for us?

I am wondering if we are covering all that we can/should? especially w.r.t. the design steps?
- what about architecture/design reviews?
- and what about retrospectives? should quality be a topic on the agenda there?
-  Also: somewhere we should discuss the role/ability to have exceptions (in a comply or explain model, I think)
- what about the impact of team organisation vs architecture, and the ability to rotate developers between teams (if all teams have a unique way of working and technology/toolset) vs the cost of changing/replacing teams? where do such topics fit?

PR #439, opened 2024-05-02
fix: handle invalid symlinks before packing for upload

nicorikken:
Check if file exist to handle invalid symlinks as they trigger an error when writing the zipfile.


nicorikken:
This was based on a situation where a repository contained a symlink in a test setup that was intended to be used in an initialised repository. In this way the error is prevented.

dennis-sig:
@nicorikken This should be the solution for the same problem as https://github.com/Software-Improvement-Group/sigridci/pull/330/files, right? Thanks a lot for the fix, though I need to check with my colleagues on the copyright statement thing.

nicorikken:
Indeed, it seems to cover the same situation. Besides the test this actually adds a fix for these cases.

dennis-sig (approved):

PR #445, opened 2024-05-21
More documentation for objectives API.

dennis-sig:
@patveck I separated the existing documentation into sections, in my view this is a bit more clear in terms of how use cases map to the available end points. I also removed the list of end point "categories", since the generated table of contents pretty much already contains that same list.

### DIFF
--- a/docs/integrations/sigrid-api-documentation.md
+++ b/docs/integrations/sigrid-api-documentation.md
@@ -518,20 +518,18 @@ This end point returns the following response structure:
       "TEST_CODE_RATIO": 0.8
     }
 
-In Sigrid, a system always has at most one objective per type. So in the above example, the 
-objective for the overall maintainability rating is apparently 4.0. The endpoint does not 
-currently indicate _why_ the maintainability target (in this example) is 4.0. There are two 
-possible reasons:
+Systems always have a single target per objective type. If you define multiple overlapping objectives, 
+precedence rules are applied to determine which objective "wins" and gets to decide the target for
+that system:
 
-- A user defined a system-level maintainability target of 4.0 stars for the given system. It 
-  doesn't matter whether there is a portfolio-wide maintainability target that the given system 
-  matches, as system-level objectives always have precedence.
-- There's no system-level maintainability target defined for the given system. However, there is at 
-  least one portfolio-wide objective that sets a maintainability target of 4.0, the given system 
-  matches the condition(s) of that objective, and there's no portfolio-wide maintainability 
-  objective with a target higher than 4.0 with conditions the given system also matches.
+- System-level objectives have precedence over portfolio-level objectives.
+- Conditional portfolio-level objectives have precedence over unconditional portfolio-level objectives.
+- Unconditional portfolio-level objectives have the lowest precedence, and therefore act like sensible
+  default policies for your portfolio.
 
-In both cases, the endpoint returns `"MAINTAINABILITY": 4.0`.
+This particular endpoint returns the target based on these precedence rules. The endpoint does not
+explain *why* that target exists. Use the [objectives status end point](#retrieving-objectives-status)
+to obtain more information on which objectives are applied to a system.
 
 #### Defining and editing system objectives
 
--- a/docs/integrations/sigrid-api-documentation.md
+++ b/docs/integrations/sigrid-api-documentation.md
@@ -47,15 +47,6 @@ Sigrid's REST API mimics this behavior, as follows:
 
 ## Available end points
 
-* [Maintainability ratings](#maintainability-ratings)
-* [Security and reliability findings](#security-and-reliability-findings)
-* [Vulnerable libraries in Open Source Health](#vulnerable-libraries-in-open-source-health)
-* [Architecture Quality ratings](#architecture-quality-ratings)
-* [Architecture Quality data](#architecture-quality-data)
-* [System metadata](#system-metadata)
-* [System lifecycle management](#system-lifecycle-management)
-* [System and portfolio objectives](#system-objectives)
-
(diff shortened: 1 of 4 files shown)


docs/integrations/sigrid-api-documentation.md
```diff
@@ -584,12 +575,65 @@ DELETE https://sigrid-says.com/rest/analysis-results/api/v1/objectives/{customer
 This would delete the system-level test code ratio objective, if it exists at the system level, and 
 return 204. In case the objective does not exist at the system level, the endpoint returns 404. 
 
+#### Retrieving objectives status
+
+    GET https://sigrid-says.com/rest/analysis-results/api/v1/objectives-evaluation/{customer}?startDate={start}&endDate={end}
+    
+This will return the current objectives status and trend for all systems in your portfolio, for the specified time period. The `startDate` and `endDate` parameters accept dates in [ISO 8601 format](https://en.wikipedia.org/wiki/ISO_8601), for example 2024-05-17.
```
patveck, line 582: I prefer "delta" instead of "trend": the former implies exactly two points in time (the start date and the end date), while the latter implies many points in time (a number of time points between the start date and end date):
```suggestion
This will return the current objectives status and delta for all systems in your portfolio, for the specified time period. The `startDate` and `endDate` parameters accept dates in [ISO 8601 format](https://en.wikipedia.org/wiki/ISO_8601), for example 2024-05-17. 
```

docs/integrations/sigrid-api-documentation.md
```diff
@@ -584,12 +575,65 @@ DELETE https://sigrid-says.com/rest/analysis-results/api/v1/objectives/{customer
...
+        }
+      ]
+    ]
+  }
+}
+```
+
+</details>
+
+The response contains every system in your portfolio, with each system containing all objectives that apply to that system. The objectives follow the same precedence rules as in Sigrid, so only objectives that actually apply will be included in the results.
```
patveck, line 623: Suggestion: try to avoid the term "apply". In principle, all portfolio objectives apply to all systems at all time, however, if the metadata of a system doesn't match the condition(s) of the objective, it doesn't result in a target.

This seems like hair splitting, but we already witnessed a case were a user thought that each portfolio objective, next to conditions, had a list of systems to which it applies (that user was asking where they could edit said list).

```suggestion
The response contains every system in your portfolio, with each system containing all objectives that apply to that system. The objectives follow the same precedence rules as in Sigrid, so only objectives with conditions that match the metadata of the system are included in the result.
```

docs/integrations/sigrid-api-documentation.md
```diff
@@ -541,6 +505,38 @@ com/rest/analysis-results/api/v1/objectives/{customer}`) is as follows:
...
+This end point returns the following response structure:
+
+    {
+      "MAINTAINABILITY": 4.0,
+      "NEW_CODE_QUALITY": 3.5,
+      "OSH_MAX_SEVERITY": "LOW",
+      "TEST_CODE_RATIO": 0.8
+    }
+
+In Sigrid, a system always has at most one objective per type. So in the above example, the
```
patveck, line 521: I probably suggested this text myself, but I'm now wondering whether it should say: "In Sigrid, a system always has at most one _target_ per type".

docs/integrations/sigrid-api-documentation.md
```diff
@@ -48,15 +48,6 @@ Sigrid's REST API mimics this behavior, as follows:
...
 
-* [Maintainability ratings](#maintainability-ratings)
-* [Security and reliability findings](#security-and-reliability-findings)
-* [Vulnerable libraries in Open Source Health](#vulnerable-libraries-in-open-source-health)
-* [Architecture Quality ratings](#architecture-quality-ratings)
-* [Architecture Quality data](#architecture-quality-data)
-* [System metadata](#system-metadata)
-* [System lifecycle management](#system-lifecycle-management)
-* [System and portfolio objectives](#system-objectives)
-
```
patveck, line 59: I'll miss this list (unless there's an auto-generated ToC).

dennis-sig: There is! The `<sig-toc>` thingy automatically creates it based on the header tags in the HTML.

docs/integrations/sigrid-api-documentation.md
dennis-sig: I rewrote the entire section a bit. I'm now separating the concept of precedence rules existing from the concrete examples. And yes, I'm intentionally not explaining the situation where you have multiple conditional objectives, I find that makes it a bit too complicated.

docs/integrations/sigrid-api-documentation.md
```diff
@@ -541,6 +505,36 @@ com/rest/analysis-results/api/v1/objectives/{customer}`) is as follows:
...
+    }
+
+Systems always have a single target per objective type. If you define multiple overlapping objectives, 
+precedence rules are applied to determine which objective "wins" and gets to decide the target for
+that system:
+
+- System-level objectives have precedence over portfolio-level objectives.
+- Conditional portfolio-level objectives have precedence over unconditional portfolio-level objectives.
+- Unconditional portfolio-level objectives have the lowest precedence, and therefore act like sensible
+  default policies for your portfolio.
```
patveck, line 528: This second bullet is what we call "Principle 2". However, it has never been implemented, so I think we should delete this second bullet entirely. Here's my suggestion:
```suggestion
- System-level objectives have precedence over portfolio-level objectives.
- If the metadata of a system matches more than one portfolio objective of the same type, the portfolio objective with the strictest target "wins". 
```

patveck (approved):

PR #452, opened 2024-06-06
Update Azure DevOps instructions for Python version and fetch depth.

dennis-sig:


MichielCuijpers (approved):
setting the cloning depth

PR #453, opened 2024-06-06
Document experimental field.

dennis-sig:


MichielCuijpers (approved):
cool to have the new option to set
experimental: true

PR #461, opened 2024-06-17
Document deleting subsystems

sbrinkhorst:
Also try to clarify when subsystems are not suitable


dennis-sig (changes requested):
Recommend to move the end point explanation, see my comment.

dennis-sig (approved):

PR #463, opened 2024-06-19
Revert "adding logos for customers idp apps"

MichielCuijpers:
Reverts Software-Improvement-Group/sigridci#462


MichielCuijpers:
by adding the logos we might confuse the customer, SIgrid only support a service-provider initiated authentication flow, so there is no need to create an OKta app with a logo

dennis-sig (approved):

PR #469, opened 2024-07-12
Update objectives.md

ShimaRafS:


dessens0:
@ShimaRafS since this duplicates a bit of what's in the guidelines, I'd suggest to move the reference to those guidelines further down to directly behind the change you just made.

PR #476, opened 2024-07-23
Add daily_findings Python example script

patveck:


examples/slack-security-findings/daily_findings.py
```diff
@@ -0,0 +1,188 @@
...
+    severity: str
+    severity_score: float
+    status: str
+
+
+class SigridApiClient:
+
+    def __init__(self, customer: str, system: str, token: str):
+        sigrid_api_endpoint = 'https://sigrid-says.com/rest/analysis-results/api/v1/security-findings/$customer/$system'
+        self.sigrid_api = Template(sigrid_api_endpoint).substitute(customer=customer, system=system)
```
dennis-sig, line 35: Wouldn't it be better to use an f-string here? AFAIK they now the preferred way of doing template strings in Python.

examples/slack-security-findings/daily_findings.py
```diff
@@ -0,0 +1,188 @@
+from __future__ import annotations
```
dennis-sig, line 1: Why did you want to enable this? AFAIK the only thing this does is move annotations to runtime instead of """compile"""-time. Unlike the other __future__ stuff, there is no Python version where the behavior is enabled by default.

examples/slack-security-findings/daily_findings.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def __init__(self, customer: str, system: str, token: str):
+        sigrid_api_endpoint = 'https://sigrid-says.com/rest/analysis-results/api/v1/security-findings/$customer/$system'
+        self.sigrid_api = Template(sigrid_api_endpoint).substitute(customer=customer, system=system)
+        self.token = token
+
+    def get_findings(self) -> Any | None:
+        try:
+            req = request.Request(self.sigrid_api)
+            req.add_header('Authorization', 'Bearer ' + self.token)
```
dennis-sig, line 41: Note this might not work if we make the token an environment variable, since environment variables are `str` and `add_header` doesn't like that.

examples/slack-security-findings/daily_findings.py
```diff
@@ -0,0 +1,188 @@
...
+    parser = ArgumentParser(description='Gets open security/reliability findings and post them to Slack.')
+    parser.add_argument('--customer', type=str, help="Name of your organization's Sigrid account.")
+    parser.add_argument('--system', type=str, help='Name of your system in Sigrid, letters/digits/hyphens only.')
+    args = parser.parse_args()
+
+    if None in [args.customer, args.system]:
+        parser.print_help()
+        sys.exit(1)
+
+    if sys.version_info.major == 2 or sys.version_info.minor < 9:
```
dennis-sig, line 168: I think this stuff is going to be repeated across every example, but I fear that's actually worth it to keep everything as self-contained and portable as possible.

examples/slack-security-findings/daily_findings.py
```diff
@@ -0,0 +1,188 @@
...
+from typing import TypedDict, Callable, Any
+from urllib import request
+from urllib.error import URLError
+import logging
+from argparse import ArgumentParser
+
+LOG = logging.getLogger(__name__)
+
+
+class Finding(TypedDict):
```
dennis-sig, line 19: `TypedDict` is a bit weird, you can happily give it fields it doesn't know, and it will happily allow it. `dataclass` is a bit more type-safe, and you can still use `asdict` to convert that into a dict. This is more pedantic, not a huge gain one way or the other.

examples/README.md
```diff
@@ -0,0 +1,6 @@
+# Sigrid integation examples
+
+This directory contains example scripts that use [Sigrid's REST API](https://docs.sigrid-says.com/integrations/sigrid-api-documentation.html). Currently, we 
+have the following examples:
+* Slack security findings: uses Sigrid's API to get open security findings for a system and
```
dennis-sig, line 5: I would make this a link.

dennis-sig (changes requested):
I have a very minor point about adding a link, all comments on the actual script are optional.

examples/slack-security-findings/daily_findings.py
patveck: Much better, changed it!

examples/slack-security-findings/daily_findings.py
patveck: I recall I needed it for `TypedDict`, which I'm no longer using.

examples/slack-security-findings/daily_findings.py
patveck: Switched over to `@dataclass`, which is nicer anyway!

examples/README.md
patveck: Done!

examples/slack-security-findings/daily_findings.py
dennis-sig: I meant in reverse, that `add_header` expects bytes which will fail when the token contains non-ASCII characters.

However, I was wrong and `add_header` happily supports `str` these days, probably for several years already.

dennis-sig (approved):

PR #481, opened 2024-07-29
Fix url with customer placeholder

tukraus:


tukraus:
Now rendered the github pages locally and the url with `<customer>` placeholder looks acceptable:
image

PR #482, opened 2024-07-29
Extend scope file schema for more advanced dependencychecker.exclude options

dennis-sig:


patveck (approved):

dennis-sig:
@fabianl-sig Could you also double-check? Without the schema change you can't actually create any scope file with the new option, because they will not pass the validation.

resources/sigrid-scope-file.schema.json
```diff
@@ -328,8 +328,51 @@
...
+                "properties": {
+                  "activity": {
+                    "type": "string"
+                  }
+                },
+                "required": ["activity"]
+              }
+            ]
+          },
+          "description": "List of paths and/or dependencies that should be ignored for Open Source Health."
```
fabianl-sig, line 375: Do we also want to update this description to include the new options and/or do we want to also add descriptions to each type?

Leaving this until it's fully rolled out is also fine by me.

test/test_scope_file_schema.py
```diff
@@ -79,3 +79,35 @@ def testDisallowedSection(self):
...
+            dependencychecker:
+              exclude:
+                - something: "noot"
+            """
+            
+        try:
+            parsedScope = yaml.load(scope, Loader=yaml.FullLoader)
+            jsonschema.validate(instance=parsedScope, schema=self.schema)
+        except jsonschema.ValidationError as e:
+            self.assertTrue("{'something': 'noot'} is not valid under any of the given schemas" in e.message)
```
fabianl-sig, line 112: If `jsonschema.ValidationError` is not thrown here the test would still pass since nothing would actually be asserted. Something like:
```suggestion
        try:
            parsedScope = yaml.load(scope, Loader=yaml.FullLoader)
            jsonschema.validate(instance=parsedScope, schema=self.schema)
            self.assertTrue(False, "ValidationError should have been raised")
        except jsonschema.ValidationError as e:
            self.assertTrue("{'something': 'noot'} is not valid under any of the given schemas" in e.message)
```
should do the trick.

fabianl-sig (changes requested):

fabianl-sig (approved):

PR #490, opened 2024-08-15
adding subsystem root info

MichielCuijpers:
adding subsystem root to the image and the 2 yaml files thanks to Hanne


MichielCuijpers:
@dessens0 I added some text to go with the image

dennis-sig (approved):

PR #495, opened 2024-09-04
Feedback: Add column to table that shows the maintainability of the code before the changes.

dennis-sig:


dennis-sig:
@patveck The funny thing is that you can use this merge request to see the change: It's the 3rd column in the "Sigrid ratings" table. The end point already has this information, we just didn't show it. This is more similar to the New/Changed Code Quality page in Sigrid, and we had some people asking if we can also add it to this feedback.

patveck (approved):

PR #498, opened 2024-09-11
Update analysis-scope-configuration.md

MichielCuijpers:


MichielCuijpers:
The componentization is not required anymore, we support automatic detection now. A missing languages should be caught by the validation service

PR #517, opened 2024-10-20
Log HTTP response (body and headers) in case of HTTPError

patveck:


sigridci/sigridci/sigrid_api_client.py
```diff
@@ -61,36 +61,55 @@ def callSigridAPI(self, path, body=None, contentType=None):
...
-                elif e.code == 410:
-                    if e.reason:
-                        UploadLog.log(f"{e.reason} (HTTP status {e.code} for {e.url})")
-                    else:
-                        UploadLog.log(f"The system no longer exists (HTTP status {e.code} for {e.url})")
-                    sys.exit(1)
-                elif allow404 and e.code == 404:
+                if e.code == 404 and allow404:
                     return False
+                self.handleError(e, server)
```
dennis-sig, line 73: Should be `SigridApiClient.handleError(…)`, since that's static. You could also make those extra methods not static, since you're never calling them in an exclusively static context anyway.

patveck: Oops! I decided to keep it static, and call it the right way.

dennis-sig (approved):

PR #520, opened 2024-10-24
Keep tests compatible with Python 3.9

dennis-sig:


sigridci/sigridci/sigrid_api_client.py
```diff
@@ -104,9 +104,11 @@ def handle401or403(e: urllib.error.HTTPError, msg: str):
 
     @staticmethod
     def printResponse(e: urllib.error.HTTPError):
-        UploadLog.log(f"Response headers:\n{str(e.headers)}")
-        if e.fp is not None:
-            UploadLog.log(f"Response body:\n{e.fp.read().decode()}")
+        UploadLog.log(f"Response headers:\n{dict(e.headers)}")
+
+        body = e.fp.read().decode("utf8") if e.fp else ""
+        if body:
```
patveck, line 110: Suppose `e.fp` is falsy, then `body` is the empty string (previous line). In all other cases, `body` is the result of `decode("utf8")`. Doesn't this imply that body is always true(-like)?

sigridci/sigridci/sigrid_api_client.py
```diff
@@ -104,9 +104,11 @@ def handle401or403(e: urllib.error.HTTPError, msg: str):
 
     @staticmethod
     def printResponse(e: urllib.error.HTTPError):
-        UploadLog.log(f"Response headers:\n{str(e.headers)}")
-        if e.fp is not None:
-            UploadLog.log(f"Response body:\n{e.fp.read().decode()}")
+        UploadLog.log(f"Response headers:\n{dict(e.headers)}")
```
patveck, line 107: If there's no body, we output "No response body". Wouldn't it be better to do the same for headers: if we detect there are 0 headers, we say "No response headers".

This is only error output, so we shouldn't over-engineer, but now it might lead to confusion, with people thinking there actually are headers, but they're somehow invisible (e.g., they start with a terminal control character setting the font color equal to the background).

sigridci/sigridci/sigrid_api_client.py
dennis-sig: An empty string is falsy, so if the response body is empty you would get the "no body" message.

sigridci/sigridci/sigrid_api_client.py
dennis-sig: Fair point. I think in a normal situation it's pretty much impossible to have no headers at all. However, we've also had "weird proxy" situations, so in that regard this *could* indeed happen. I'll change it.

patveck (approved):

PR #521, opened 2024-10-25
Update documentation for recent Sigrid UI changes.

dennis-sig:


docs/reference/glossary.md
ShimaRafS, line 1: I recall at some point we wanted to add the term "Application" also as anther way to call "system", is that still an interest?

ShimaRafS (approved):

PR #522, opened 2024-10-25
Update user management and sso documentation with open id connect instructions

valdore86:


valdore86:
I've not yet added any examples, but this should fulfill the first need for some OIDC documentation.

dennis-sig (approved):

PR #530, opened 2024-11-05
Make Sigrid CI feedback more visually appealing.

dennis-sig:
Multiple people have told us that the Sigrid CI comments are a bit too large vertically. We made several changes based on this feedback: 

- If there are more than 8 refactoring candidates, we limit the list and just show "…and 8 more". Showing every single refactoring candidate for large files becomes overwhelming.
- For the remaining/unchanged technical debt, we remove the list altogether and instead have a link to Sigrid. This list can be *really* long if your system has a lot of technical debt, so this is best explored with the Sigrid UI since such a long list is not browsable.
- We put all the details beyond the conclusion in ` <summary>`/`<details>` block. Unfortunately not every platform supports this, as they all have slightly different flavors of Markdown. To account for this, we'll only add these constructs if we know the platform supports them.
- Put the Sigrid link in the title, instead of having to hunt for it all the way down.
- Duplication refactoring candidates were extremely verbose if there are many duplicate occurrences, this is now displayed a bit more elegantly.
- Some general UX tweaks by Yu.


sigridci/sigridci/platform.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class Platform:
+    @staticmethod
+    def isGitHub():
+        return "GITHUB_REPOSITORY" in os.environ
+
+    @staticmethod
+    def isGitLab():
+        return "CI_PROJECT_PATH" in os.environ
```
patveck, line 25: We could use variable `GITLAB_CI` instead. The name `CI_PROJECT_PATH` is rather generic, there might be other environments that also use it. On GitLab, `GITLAB_CI` is defined and typically has the value `true`.

test/test_markdown_report.py
```diff
@@ -145,6 +143,43 @@ def testSortRefactoringCandidatesTableBySeverity(self):
...
+            |------|-----------------|----------|
+            | 🟠 | **Unit Size**<br />(Introduced) | aap-1 |
+            | 🟠 | **Unit Size**<br />(Introduced) | aap-2 |
+            | 🟠 | **Unit Size**<br />(Introduced) | aap-3 |
+            | 🟠 | **Unit Size**<br />(Introduced) | aap-4 |
+            | 🟠 | **Unit Size**<br />(Introduced) | aap-5 |
+            | 🟠 | **Unit Size**<br />(Introduced) | aap-6 |
+            | 🟠 | **Unit Size**<br />(Introduced) | aap-7 |
+            | 🟠 | **Unit Size**<br />(Introduced) | aap-8 |
+            | ⚫️ | | + 91 more |
```
patveck, line 163: I'm reading that `range` in Python is exclusive (we're generating 99 refactoring candidates, not 100). In that case, this is correct, but it is a bit weird. Probably nothing we can do about it.

dennis-sig: It's because I start at 1 instead of 0. So the `range(1, 100)` will include everything between 1 and 99, which is indeed 99 elements.

sigridci/sigridci/platform.py
dennis-sig: Fair point, changed. By that logic the Azure one is maybe also a bit generic, but from briefly looking at their list I have the feeling they're *all* generic like that.

patveck (approved):

PR #531, opened 2024-11-05
Clarification subsystem configuration

syllosigm:
Based on recent ticket


MichielCuijpers (approved):
thanks that make it more clear

PR #539, opened 2024-11-13
Gitlab update comment

dennis-sig:


test/test_gitlab_pull_request_report.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __init__(self, existingCommentId):
+        super().__init__()
+        self.existingCommentId = existingCommentId
+
+    def buildPostCommentURL(self, existingCommentId):
+        return str(existingCommentId)
+
+    def callAPI(self, method, url, body):
+        return self
```
patveck, line 71: Suggestion: make `callAPI` store the method (and maybe also the URL) in a class-level variable. That way, the test can verify an expectation, for instance verify that `PUT` was used in some cases and `POST` in others.

patveck (changes requested):

test/test_gitlab_pull_request_report.py
dennis-sig: Good suggestion, extended the tests.

test/test_api_caller.py
```diff
@@ -0,0 +1,65 @@
...
+        api.handleError(error)
+
+        expectedLog = [
+            "HTTP Error 404: No reason",
+            "No response headers",
+            "No response body"
+        ]
+
+        self.assertEqual(expectedLog, UploadLog.history)
+
```
patveck, line 65: Do we want to move the tests for the other error conditions here as well (`TimeOutError`, `URLerror`, etc.)?

dennis-sig: Fair point, moved the tests.

patveck (approved):

PR #545, opened 2024-11-21
Update upload-instructions.md - fix incorrect zip command example

florislw:
The old command:
zip -d code code.zip
is incorrect. It seems the -d flag is there to remove the code directory, however this flag is used to remove directories from existing zip files.

The new command:
zip -r code.zip .
recursively zips all files in the current directory. This means the code directory that is created while pulling is not deleted. We discussed in the SCC team that this is fine. We assume people know how to remove a directory from their computer. This keeps the example simple and therefore more easy to understand.

The new command does not create an unnecessary top level directory inside the zip file.


dennis-sig (approved):

dennis-sig:
Thanks! I can't tell you how that original command came to be, the manual upload instructions have been around for a *long* time. But this is definitely an improvement.

PR #546, opened 2024-11-22
Show example on how to get the scope file contents using the Sigrid API.

dennis-sig:


examples/get-scope-file/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+- The script requires Python 3.9+.
+- You will need a valid [API token](https://docs.sigrid-says.com/organization-integration/authentication-tokens.html) to access the [Sigrid REST API](https://docs.sigrid-says.com/integrations/sigrid-api-documentation.html).
+- Your API token should be available to the script as the environment variable `SIGRID_CI_TOKEN`.
+
+## Usage
+
+Once all prerequisites are in place, you can use the script:
+
+    ./get_sigrid_scope_file.py --customer mycustomername --system mysystemname
```
tukraus, line 22: Can we mark this as a code block for readability, and indicate which are the placeholders?
```suggestion
```
    ./get_sigrid_scope_file.py --customer <mycustomername> --system <mysystemname>
```
```

tukraus: Ah, I didn't see that it's already a code block due to 4-space indent. In other examples we also don't mark the placeholders with `< >` so we could leave that here as well.

PR #555, opened 2024-12-05
Sigrid CI feedback for Security and Open Source Health (only on-premise Sigrid for now)

dennis-sig:
As discussed, this splits Sigrid CI into multiple entry points, which would correspond to different steps in the pipeline:

- The "main" sigridci.py checks the input, validates the scope file, creates the upload, sends the upload, starts the analysis.
- sigridci_osh.py and sigridci_security.py are new entry points that poll the Sigrid API until the analysis results are available, and then provide feedback based on the analysis results. 
    - The reason these are *separate* entry points, is to allow situations where people want to fail a step on OSH but not on Security. If we would combine everything into one large feedback, it would be everything-or-nothing. 
    - This approach is not entirely symmetrical, because the initial entry point *also* provides maintainability feedback. This is for backwards compatibility, changing that behavior would require every single Sigrid CI user to update their pipeline configuration. We will probably make maintainability feedback optional in a future version, but for now it's required (also note that in Sigrid itself maintainability is also always required, considering the whole maintainability-is-the-foundation-of-everything story).
    - In the on-premise version, we feed the analysis results to the entry points. This part is already available.
    - For the SaaS version, we need to pass the analysis ID between these entry points. Every CI environment has a way of passing variables between steps. However, we took the lazy way out and just write analysisid.txt to the output directory. This has two benefits: We don't need to document the instructions for each environment, and it means less configuration and hassle for people wanting to add Sigrid CI to their pipeline.
    - Actually polling the Sigrid API for the analysis results will trigger an error, as those end points don't actually exist yet. 

Both OSH and Security report feedback against your Sigrid objectives. If you didn't set an explicit objective, it checks against a default objective. This is consistent with maintainability where we use a default objective of 3.5 stars.

(And I created a sub-package for "reports", which is the different ways you can consume output. The list of reports used to be extremely straightforward, but over time we added more and more. I'm mentioning this purely because it explains why so many files are touched.)

Despite the end points not being available, I would suggest we already have a look at merging the new entry points, since it would allow us to enable Sigrid CI for Security and OSH in the on-premise version. The on-premise documentation has been updated accordingly. We can then add calling the end points in a follow-up, and combine that with documenting the instructions for the SaaS version.


test/test_maintainability_markdown_report.py
```diff
@@ -44,14 +45,13 @@ def testMarkdown(self):
...
-        report = MarkdownReport()
-        report.ALLOW_FANCY_MARKDOWN = False
+        report = MaintainabilityMarkdownReport()
         markdown = report.renderMarkdown("1234", feedback, self.options)
 
         expected = """
             # [Sigrid](https://sigrid-says.com/aap/noot) maintainability feedback
             
-            **↗️  You improved your code's maintainability towards your Sigrid objective of 3.5 stars**
+            **↗️  You improved your code's maintainability towards your objective of 3.5 stars**
```
patveck, line 54: While this possessive is correct from a grammatical perspective, I wonder whether we really need it:
```suggestion
            **↗️  You improved maintainability towards your objective of 3.5 stars**
```

Or do you literally mean: maintainability of the code that **_you_** wrote has been improved? So not all code, but yours?

dennis-sig: Somehow *"You improved maintainability"* feels like there is a word missing. Maybe @olimoli9160 wants to weigh in as a native speaker?

sigridci/sigridci_maintainability.py
```diff
@@ -0,0 +1,44 @@
...
+
+    with open(args.analysisresults, mode="r", encoding="utf-8") as f:
+        analysisId = "local"
+        feedback = json.load(f)
+
+    markdownReport = MaintainabilityMarkdownReport()
+    additionalReports = [GitLabPullRequestReport(markdownReport), AzurePullRequestReport(markdownReport)]
+
+    runner = SigridCiRunner(options, apiClient)
+    runner.provideFeedbackOnly(analysisId, feedback, markdownReport, additionalReports)
```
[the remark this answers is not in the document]
dennis-sig: Fair point. The whole "just give me feedback" is basically a separate path, so now that gets its own `FeedbackProvider`. Also added test for that thing.

sigridci/sigridci_osh.py
```diff
@@ -0,0 +1,43 @@
...
+
+from sigridci.command_line_helper import getFeedbackPublishOptions, parseFeedbackCommandLineArguments, checkEnvironment
+from sigridci.reports.azure_pull_request_report import AzurePullRequestReport
+from sigridci.reports.gitlab_pull_request_report import GitLabPullRequestReport
+from sigridci.reports.osh_markdown_report import OpenSourceHealthMarkdownReport
+from sigridci.sigrid_api_client import SigridApiClient
+from sigridci.sigridci_runner import SigridCiRunner
+
+
+if __name__ == "__main__":
```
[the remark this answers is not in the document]
dennis-sig: I don't think we will ever have different flows for the different capabilities, so maybe it doesn't even make sense to give them all their own entry point. Also doesn't scale particularly well if we also want to offer e.g. Reliability in the future. So I went back to a single entry point with a parameter to select a capability. The new entry point is now hooked up to a separate flow, based on what you suggested above.

test/test_maintainability_markdown_report.py
olimoli9160: I would say "You improved the maintainability of the code/codebase/system..."  or "You improved the codebase's maintainability..." to remove the personal possessiveness of the phrasing while keeping it in line with natural language. 

I'm not going to pretend to be an english scholar, but my thinking is the following - Maintainability in this case is the target of the subject, but the subject is compound because it describes a quality of the secondary noun, that being the code itself. So it reads odd to me to leave out secondary noun here. It's not necessarily grammatically wrong to say "you improved maintainability" and leave it at that, but it doesn't flow naturally when I read it back.

dennis-sig: With that in mind, would you still phrase it as "your objective"? Right now the text is like this:

- You wrote maintainable code and achieved your objective of X
- You improved your code's maintainability towards your objective of X
- Your rating did not change and is still below your objective of X
- Your code did not improve towards your objective of X
- You did not change any files that are measured by Sigrid

dennis-sig: (Rephrased the summary text based on the suggestion by Michael)

patveck (approved):

patveck (approved):

PR #557, opened 2024-12-09
Add gitconfig to images to disable safe directory checking

dessens0:
In our documentation, we describe an [issue](https://docs.sigrid-says.com/capabilities/faq.html#im-receiving-an-error-message-about-dubious-ownership-in-repository) that people can run into when Sigrid CI tries to create a commit history digest. By adding a pre-defined `.gitconfig` to our images, this issue should not longer occur for user who use our Docker images to run Sigrid CI.

### DIFF
--- a/sigridci/gitconfig-add-safe-directory
+++ b/sigridci/gitconfig-add-safe-directory
@@ -1,0 +1,2 @@
+[safe]
+	directory = *
--- a/sigridci/gitconfig-add-safe-directory
+++ b/sigridci/gitconfig-add-safe-directory
@@ -1,2 +1,0 @@
-[safe]
-	directory = *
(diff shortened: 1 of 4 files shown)


sigridci/gitconfig-add-safe-directory
```diff
@@ -0,0 +1,2 @@
+[safe]
+	directory = *
```
tukraus, line 2: This line now seems to start with a tab character instead of 4 spaces, was that intentional? I'm not entirely sure if that just works? Out of curiosity, why not just running the `git config --global --add safe.directory '*'` command so it will generate a config file that it will like for sure?

dessens0: Ah yes, that's better! The reason I arrived at this solution is [this post](https://stackoverflow.com/a/71904131) that suggested some issues with running the command on different platforms, but that was when I still had the idea to run the command in the Python script 😬

tukraus (approved):

PR #570, opened 2024-12-18
Use new on-premise container names.

dennis-sig:


docs/organization-integration/onpremise-analysis.md
```diff
@@ -135,8 +135,10 @@ The following example shows how to start an ad-hoc analysis for a system located
     docker run \
       --env-file sigrid-ci-config.txt \
       -v /mysystem:/code \
-      -ti softwareimprovementgroup/sigrid-multi-analysis:1.0.20241206 \
+      -ti softwareimprovementgroup/sigrid-multi-analyzer:1.0.20241206 \
```
patveck, line 138: Should we repeat in this section about manually running the analyzers that this requires logging in to Docker and that on-prem customers get the credentials from SIG? And even repeat it again in the part at the beginning of this file about running this in CI (it does say something in a comment)?

dennis-sig: I added references to the corresponding section. I would prefer to avoid a paragraph of text, also to avoid the repetition.

patveck (approved):

PR #573, opened 2024-12-18
Support proxy configuration via environment variable.

dennis-sig:
@patveck This is a backport of something that was shared with us by a client. We're using an environment variable rather than a command line flag, as (A) this is actually part of the environment (B) this allows people to pass the proxy as a secret.


patveck (approved):

dennis-sig:
@MichielCuijpers FYI.

PR #574, opened 2024-12-18
Fix file path

patveck:


docs/organization-integration/onpremise-analysis.md
```diff
@@ -136,7 +136,7 @@ The following example shows how to start an ad-hoc analysis for a system located
 
     docker run \
       --env-file sigrid-ci-config.txt \
-      -v /mysystem:/code \
+      -v /mysystem:/tmp/sources \
```
patveck, line 139: @dennis-sig I checked: in case the sigrid-multi-analyzer is not running in a supported CI/CD environment, it expects the root of the sources at `/tmp/sources`.

In addition, `/mysystem` is a local directory. Few corporate users will have write access at the root level, and on Windows, it is probably completely different. Do you think we need to make a note about that?

dennis-sig: `/mysystem` was intended as a random dummy value. I'm fine with changing it into something else, but wouldn't you have that issue with any non-real example?

patveck: Yes, let's leave it as-is for that reason. Also, this is basic Docker knowledge.

PR #579, opened 2025-01-14
Configure GitHub Actions permissions in workflow

praseodym:
This also works for organization accounts where the 'Workflow permissions' setting cannot be changed, e.g.:
Screenshot 2025-01-14 at 09 37 17


MichielCuijpers:
thank you @praseodym for taking the time, we will look at your PR.

patveck (approved):
Thanks, valuable PR! I'll merge as soon as I get the pipeline jobs to pass (they currently fail, but that seems entirely unrelated to the contents of this PR).

PR #581, opened 2025-01-15
Update action.yaml

bruno-oliveira:
This ensures that the convert argument gets passed as an argument to the underlying docker image to run Sigrid CI

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -114,6 +114,8 @@ runs:
     - ${{ inputs.include }}
     - "--subsystem"
     - ${{ inputs.subsystem }}
+    - "--convert"
+    - ${{ inputs.convert }}
   env:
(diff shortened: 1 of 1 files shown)


MichielCuijpers (approved):
this was missing indeed

action.yaml
```diff
@@ -114,6 +114,8 @@ runs:
     - ${{ inputs.include }}
     - "--subsystem"
     - ${{ inputs.subsystem }}
+    - "--convert"
+    - ${{ inputs.convert }}
```
patveck, line 118: The `convert` input is not required, see line 48. This raises the question: what would the command line look like if the user indeed does not provide a value for `convert`. Would GitHub add just `--convert` to the command line? If so, how would the Python script deal with that?

PR #584, opened 2025-01-20
Describe rate limiting thresholds

tukraus:
- Update generic request rate limit (back to original 2500)
- Introducing analysis request rate limiting.


MichielCuijpers (approved):
ok

PR #604, opened 2025-02-05
Empty scope file should still be validated.

dennis-sig:


test/test_sigridci_runner.py
```diff
@@ -343,6 +343,20 @@ def testFailIfScopeFileIsNotValid(self):
 
         self.assertEqual(UploadLog.history, expectedLog)
 
+    def testEmptyScopeFileIsNotValid(self):
```
Geertien, line 346: For completion, can you add a test for the metadata file as well? (That should exit as well right?) It looks a bit redundant as it is almost the same test, but I think it's good to have both checked as it's technically separate logic.

Geertien (changes requested):
Looks good, I'd only like an additional test 🙂

test/test_sigridci_runner.py
dennis-sig: OK.

dennis-sig: Added. While making the test I found it interesting that the URLs for validating the scope file and metadata file are so different, but let's leave that out of this ticket.

Geertien: Looks good 🙂

Geertien (approved):
Nice. Well found by the way!

PR #611, opened 2025-02-11
Change how we run Sigrid CI on itself.

dennis-sig:
We previously excludes the `sigridci` directory from the upload, as some people clone it *within* their own repository. That creates a problem for ourselves though, since it means we could not run Sigrid CI on the codebase for Sigrid CI. 

Proposal:

- We *do* include the directory in the upload.
- Nowadays the `sigridci` directory is part of the default excludes, so it won't be analyzed.
- We then switch off the default excludes for ourselves.
- We no longer need to rename the directory when uploading the code.

### DIFF
--- a/.github/workflows/sigrid-publish.yml
+++ b/.github/workflows/sigrid-publish.yml
@@ -10,9 +10,7 @@ jobs:
     steps:
       - name: "Check out repository"
         uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


Geertien (approved):
The changes look good to me 🙂 

Technically it could be of course that there are customers who also have `default_excludes: false` AND have the sigridci directory in their codebase (because of security reasons as described in the documentation), but that is relatively unlikely and if they do exist they can still adjust their scope to explicitly exclude it 🙂

PR #618, opened 2025-02-17
Hard exclude our own Dockerfile from the upload.

dennis-sig:


Geertien (approved):
This is fair for now 🙂

PR #619, opened 2025-02-18
Improve styles for narrow viewports

florislw:
This PR fixes two issues on narrow viewports.
1. The article did not have horizontal margin (implemented with 20px padding in css and increasing max-width of article by 40px).
2. The top bar design broke on smaller screens. Quick fix for now is making it scrollable.

### DIFF
--- a/docs/_includes/stylesheet.css
+++ b/docs/_includes/stylesheet.css
@@ -109,6 +109,8 @@ ol li::before {
     position: sticky;
     z-index: 2;
     top: 0;
+    overflow-x: auto;
+    overflow-y: hidden;
 }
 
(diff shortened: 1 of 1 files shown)


dennis-sig:
@florislw Thanks! We're not going full responsive with the documentation, as less than 0.1% of people use it on mobile/tablet, and we need to prioritize where we spend our time. This is obviously better though, so we can of course still do this one.

dennis-sig (approved):

florislw:
@dennis-sig I understand. I had it open on my vertical screen and saw these issues, quick fix. Glad It is appreciated :)

PR #620, opened 2025-02-19
Revert "Change how we run Sigrid CI on itself."

dennis-sig:
Reverts Software-Improvement-Group/sigridci#611


Geertien (approved):
It's a shame for us it's needed, but not too problematic and the best course of action for now indeed. I don't see anything weird 🙂

PR #622, opened 2025-02-19
Show error message when trying to use a scope file for a subsystem.

dennis-sig:


sigridci/sigridci/sigridci_runner.py
```diff
@@ -96,6 +96,12 @@ def displayFeedback(self, analysisId):
     def validateConfigurationFiles(self):
         scope = self.options.readScopeFile()
         if scope is not None:
+            if self.options.subsystem not in (None, "", "root", "scopefile"):
+                UploadLog.log("-" * 80)
+                UploadLog.log("You cannot provide a scope configuration file for a subsystem.")
```
sbrinkhorst, line 101: I think we should be a bit clearer about what this means in practice. Because now we say that you can't do something but it doesn't fail, so you clearly can do it. It is just that we don't process it. 

"Scope configuration files are not applicable to subsystems and will be ignored"

dennis-sig: Good one. Do you then recommend to log this message and proceed? Or should we block them from proceeding?

sbrinkhorst: Log this message and proceed.

dennis-sig: Like this? I think it's fair to call it a warning if we're not going to kick people out.

sbrinkhorst (approved):
Yes, looks good. I trust you on the documentation changes.

PR #623, opened 2025-02-24
Handle links within a subsystem.

dennis-sig:
When you're using sub-systems, the file paths you see in Sigrid do not correspond to what you see in the repository. This removes the sub-system prefix from the generated link, so the file path matches the repository.

Somewhat confusingly, the *label* is left unchanged, since those labels still reflect what you see in Sigrid. But the least we can do is trying to avoid giving people a broken link, so I still think this makes sense.

### DIFF
--- a/test/test_maintainability_markdown_report.py
+++ b/test/test_maintainability_markdown_report.py
@@ -134,7 +134,7 @@ class MarkdownReportTest(TestCase):
 
         report = MaintainabilityMarkdownReport()
         report.decorateLinks = False
-        table = report.renderRefactoringCandidatesTable(refactoringCandidates)
+        table = report.renderRefactoringCandidatesTable(refactoringCandidates, self.options)
 
         expected = """
             | Risk | System property | Location |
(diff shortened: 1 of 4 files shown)


test/test_maintainability_markdown_report.py
```diff
@@ -610,6 +610,28 @@ def testMakeLinksForUnitMetrics(self):
 
         self.assertEqual(table.strip(), inspect.cleandoc(expected).strip())
 
+    @mock.patch.dict(os.environ, {
+        "CI_SERVER_URL" : "https://example.com",
+        "CI_PROJECT_PATH" : "aap/noot",
+        "CI_COMMIT_REF_NAME" : "mybranch"
+    })
+    def testNormalizeLinksForSubsystems(self):
+        rc = self.toRefactoringCandidate(f"jan/aap.py::noot", "introduced", "UNIT_SIZE", "VERY_HIGH")
```
patveck, line 619: Pedantic: `jan` is not part of our standard set of test data names. Consider `jet`, `mies` or `henk`. My first reaction now was: what's (our) Jan got to do with this.

patveck (changes requested):

test/test_maintainability_markdown_report.py
dennis-sig: Changed it. But I did already previously use `jan` as a test name as an in-joke, which has already been discovered by the real @janlaan :)

patveck (approved):

PR #627, opened 2025-03-07
Improve error message when validating system name

janlaan:
`PublishOptions.isValidSystemName()` contains a check for `isDigit()`, but this is not present in the error message when a system name is invalid. This causes potential confusion. This change should make sure the error message covers all error scenarios.


dennis-sig:
Addressed by https://github.com/Software-Improvement-Group/sigridci/pull/628

PR #633, opened 2025-03-21
Document usage of custom certificates in the helm chart and analysis container

fonteijnf:


docs/organization-integration/onpremise-kubernetes.md
```diff
@@ -496,6 +512,29 @@ inbound-api:
...
+### (F.3) Optional: Configure Custom Certificates
+When using a self-signed certificate for the S3-compatible storage we also have to tell inbound-api which cert specifically should be used for S3 so that it can also share it with the importer jobs it starts.
+```yaml
+inbound-api:
+  config:
+    importJob:
+      objectStoreCertificate:
+        # -- If the object-storage (S3) requires a custom certificate configure the custom certificates section. Then
+        # set key to the filename of the specific S3 certificate.
+        key: ""
```
fabianl-sig, line 524: Do we want to add the example key or leave it like this.
```suggestion
        key: my-object-store-cert.pem
```

fabianl-sig (approved):

PR #636, opened 2025-03-25
Sigrid CI: Show fixed refactoring candidates in table.

dennis-sig:
@Geertien I forgot we need to actually start displaying the fixed refactoring candidates in the "what went well" section, otherwise you can't actually see them. I think we can review/merge this once the back-end changes are ready.

(Note: We round the release notes to the first day of the sprint, so you can occasionally get dates slightly into the future.)

### DIFF
--- a/test/test_ascii_art_report.py
+++ b/test/test_ascii_art_report.py
@@ -52,9 +52,9 @@ class AsciiArtReportTest(TestCase):
         
         report = AsciiArtReport()
         
-        self.assertEqual(report.formatRefactoringCandidate(rc1), "    - (introduced)   aap")
-        self.assertEqual(report.formatRefactoringCandidate(rc2), "    - (worsened)     noot\n                     mies")
-        self.assertEqual(report.formatRefactoringCandidate(rc3), "    - (worsened)     noot\n                     mies")
+        self.assertEqual(report.formatRefactoringCandidate(rc1), "Unit Size (introduced)\n  aap")
+        self.assertEqual(report.formatRefactoringCandidate(rc2), "Duplication (worsened)\n  noot\n  mies")
+        self.assertEqual(report.formatRefactoringCandidate(rc3), "Unit Size (worsened)\n  noot\n  mies")
 
     def testPrintRegularReport(self):
         feedback = {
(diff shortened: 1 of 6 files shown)


Geertien (approved):
These changes look good.

As discussed separately however, the new fixed. refactoring candidates will already show up in the ascii art report, but just mixed in with the rest. We might want to consider changing this behaviour as well. Some options are:
- Replicate the "what went well" / "what could be better" / "Remaining technical debt (number)" split to the ascii art
- Do something with sorting the refactoring candidates
- Show fixed refactoring candidates in another color (e.g. green)

It might make sense to do that in a separate PR though, I think that would be fine as already without this PR they'll show up in the ascii art now. So I'll approve but you may choose to add it to this PR still 🙂

test/test_ascii_art_report.py
```diff
@@ -72,28 +76,35 @@ def testPrintRegularReport(self):
...
-            Unit Size
-                - (introduced)   a.py
-                                 aap()
+            -------------------------------------------------------------------------------
+            What could be better?
+            -------------------------------------------------------------------------------
+            
+            Unit Size (introduced)
+              a.py
+              aap()
```
Geertien, line 92: Before reading the actual code I thought there was still categorisation on the metric-category combination and under that the refactoring candidates would be displayed. Do I understand correctly that instead every refactoring candidate will get these three lines followed by a white line? While I don't necessarily have a problem with this representation as such I do note that it effectively doubles the number of lines the RC list takes in the output. Especially for larger MRs/PRs this can cause the list to be very long. I'm not sure that's desirable. Have you considered this?

Also separately, I noticed that the original representation was still made without the risk categories in mind (I think they were not supplied yet). Now that we have them and even use them to sort the candidates, do we maybe also want to display them in the text output?

Geertien (commented):
Nice further improvements. I think it in general looks good, but there's two further considerations I added (in one comment). I'm not entirely sure that they should change, but they should at least be considered before merging.

test/test_ascii_art_report.py
Geertien: Yeah I agree with those assertions. As discussed earlier I had in mind to have further sub-categorisation under the main categories, but that would indeed make it overly complex and I hadn't considered that we stopped explicitly displaying unchanged candidates which saves lines again. I think the current feedback looks good and is quite the improvement. If needed we can also always reconsider based upon feedback of course.

Geertien (approved):
Looks all good now. Some nice iterations leading to a very good improvement I believe 🙂

PR #639, opened 2025-03-26
Add column to ASCII art feedback to match what we see in Markdown.

dennis-sig:


MichielCuijpers:
thanks !

sigridci/sigridci/reports/ascii_art_report.py
```diff
@@ -24,36 +24,42 @@ class AsciiArtReport(Report):
...
     ANSI_BLUE = "\033[96m"
-    LINE_WIDTH = 70
+    LINE_WIDTH = 80
 
     def __init__(self, output=sys.stdout, ansiColors=True):
         self.output = output
         self.ansiColors = ansiColors
 
     def generate(self, analysisId, feedback, options):
-        publish = options.runMode != RunMode.FEEDBACK_ONLY
```
Geertien, line 34: This was just a leftover from when we used to change the exact feedback/columns based on this right? (seems from a quick look back) Good to clean it up then, but lest directly also clean up the import line `from ..publish_options import RunMode` then 🙂

sigridci/sigridci/reports/ascii_art_report.py
```diff
@@ -24,36 +24,42 @@ class AsciiArtReport(Report):
...
                 self.formatRating(feedback["newCodeRatings"], metric)
-            ]
-
-            self.printTableRow(row)
+            )
 
-    def printTableRow(self, row):
-        formattedRow = "%-27s%-25s%-18s" % tuple(row)
+    def printTableRow(self, *row):
+        formattedRow = "%-25s%-22s%-17s%-22s" % tuple(row)
```
Geertien, line 62: This adds up to `86` not `80`. I noticed something was off from the extra whitespace added outside the dashed line in the test differences. I suppose that unnecessarily causes line wrapping still if there's an actual line width of 80 used in the console. It's just only the extra whitespace at the end though, the text ends at 80 now so we can just fix it by reducing the last column.

Still when I was looking I got a bit bothered by there being two spaces between the middle two column headers and three between the right two columns. I see though that 3 aligns with the spaces after "Component Independence" and "Component Entanglement", although I think that's slightly different from having equal spacing between the headers and it might be desirable to keep it at 3. Anyway this isn't that important, but maybe as you're changing the numbers anyway you can also quickly try aligning all columns or the rating columns to have 2 spaces between them. I'll agree with any option as long as it actually adds up to 80 (or less).

Geertien (changes requested):
Logic-wise it all looks good to me. Just two minor remarks to address 🙂

sigridci/sigridci/reports/ascii_art_report.py
dennis-sig: Yeah the IDE informed me it wasn't used anymore. I removed the import.

Geertien: Nicely cleaned up 🙂

sigridci/sigridci/reports/ascii_art_report.py
Geertien: Looks good 🙂

Geertien (approved):
Looks all nice and good to me now!

PR #640, opened 2025-03-26
Make documentation responsive.

dennis-sig:


Cheer-ding (approved):
- Responsive V1
- Body font weight changed into 300

PR #642, opened 2025-03-28
Reduce header size in mobile version.

dennis-sig:


Cheer-ding (approved):

Cheer-ding:
- Mobile H1 size reduce to 24px;
- Body line height from 1.3 to 1.6;
- Strong font from 500 to 400;

PR #651, opened 2025-04-14
Report on Jenkins being used.

dennis-sig:
The standard environment variables in Jenkins have super generic names. BUILD_NUMBER seems like the one I consider least likely to collide within something else.

### DIFF
--- a/sigridci/sigridci/platform.py
+++ b/sigridci/sigridci/platform.py
@@ -34,6 +34,10 @@ class Platform:
         return "BITBUCKET_REPO_SLUG" in os.environ
 
     @staticmethod
+    def isJenkins():
+        return "BUILD_NUMBER" in os.environ
+
+    @staticmethod
     def getPlatformId():
         if Platform.isGitHub():
             return "github"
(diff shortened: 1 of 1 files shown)


sigridci/sigridci/platform.py
```diff
@@ -43,6 +47,8 @@ def getPlatformId():
             return "azure-devops"
         elif Platform.isBitBucket():
             return "bitbucket"
+        elif Platform.isJenkins():
```
patveck, line 50: Just a remark: this method looks for `BUILD_NUMBER`. That's rather generic: I wouldn't be surprised if one of the other platforms als sets that exact same env var (now or in the future). Consequently, order is important here: if we discover `BUILD_NUMBER` but we haven't concluded it is any of the other platforms, then we conclude it must be Jenkins. That's probably exactly what we want, but we might want to add a comment to remind our future selves of that.

patveck (approved):

sigridci/sigridci/platform.py
dennis-sig: Good one, added a comment. It was indeed intentional to check Jenkins last, so future platforms should be checked *before* Jenkins.

patveck (approved):

PR #652, opened 2025-04-16
Add documentation for new categories property for security and reliab…

patveck:
…ility endpoints


dennis-sig:
@patveck I find this noteworthy enough to reference this in the release notes.

dennis-sig (approved):

PR #656, opened 2025-05-07
Adding m2/ as variant

MichielCuijpers:
we noticed a customer where the maven cache was in a m2/repo instead of the .m2/repo. This PR will add that variant

### DIFF
--- a/sigridci/sigridci/system_upload_packer.py
+++ b/sigridci/sigridci/system_upload_packer.py
@@ -46,6 +46,7 @@ class SystemUploadPacker:
         ".gradle/",
         ".idea/",
         ".m2/",
+        "m2/"
         ".terraform/",
         ".yarn/"
     ]
--- a/sigridci/sigridci/system_upload_packer.py
+++ b/sigridci/sigridci/system_upload_packer.py
@@ -46,7 +46,7 @@ class SystemUploadPacker:
         ".gradle/",
         ".idea/",
         ".m2/",
-        "m2/"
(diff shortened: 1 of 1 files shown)


Geertien:
Although this change would indeed solve the problem at hand, it also poses a risk to exclude genuine directories. Given our current implementation this would also exclude any directories with names ending with "m2" (e.g. `team2`), while to be analysed directories fully named `m2` are unlikely, directory names ending in "m2" aren't necessarily. We could consider making the exclude `/m2/`, that might not work if the directory is at the root of the filesystem, but that's probably an unlikely edge case. Alternatively we could change the current implementation, but that is a bigger investment and probably not reasonable for this case.

I do wonder though whether it makes much sense to make a default exclude for this at all. If I'm not mistaken, the only way of moving away from the standard of `${user.home}/.m2/repository/` is to explicitly configure another Maven Local Repository (see [documentation](https://maven.apache.org/guides/mini/guide-configuring-maven.html)), in this case it was apparently chosen to use `[some path to the upload root]/m2/repo`, but there is no limitation to use `m2` in any directory name in that path, so customers could use any directory name and this extra exclude only helps for this specific case for this specific customer.

So in short, I'd suggest to not add this default exclude at all and instead ask the customer to add `"m2/"` or `"/m2/"` to their manual exclude patterns. Or if that's not desirable change the default exclude to `"/m2/"`, assuming that works for this case.

dennis-sig (approved):

PR #659, opened 2025-05-12
Make exclude more focused.

dennis-sig:


Geertien (approved):
I think this decreases the likeliness of catching directories that we do actually want to upload sufficiently 🙂

PR #667, opened 2025-05-28
Use on-premise SSL context if available.

dennis-sig:


sigridci/sigridci/sigrid_api_client.py
```diff
@@ -36,6 +37,10 @@ def __init__(self, options: PublishOptions):
         self.urlCustomerName = urllib.parse.quote_plus(options.customer.lower())
         self.urlSystemName = urllib.parse.quote_plus(options.system.lower())
         self.token = os.environ["SIGRID_CI_TOKEN"].strip()
+        if os.environ.get("SIGRID_CA_CERT"):
```
dessens0, line 40: [nitpick] I suppose the Pythonic way to write this is
```Python
self.sslContext = ssl.create_default_context(cafile=os.getenv("SIGRID_CA_CERT"))
    if os.environ.get("SIGRID_CA_CERT") else None
```

dennis-sig: I didn't use the ternary because then the line went over 120 characters, but I also don't have objections, so added.

PR #679, opened 2025-06-13
Update documentation for new features.

dennis-sig:


docs/capabilities/reports/modernization-report.md
rgopal-sig, line 1: I think the documentation can benefit from clarifying that modernization scenarios are _proposed_. Currently it is a bit confusing whether this requires user input on scenario selection.

In the blue column explanation, the scenario bullet could have "the proposed scenario" instead of "the chosen scenario". Similar for the 'Deciding on modernization scenarios'. That could be clearer with "Generating modernization scenarios".

There's a space in the word 'modernization' underneat the table-image.

In the bubble chart, it's not specified what the bubble size indicates. The X-asis specifies the modernization execution effort I believe. The current description doesn't make that clear.

The tracking of the modernization scenarios using Sigrid objectives should have an example and  the clarification that the objectives only relate to the technical quality objectives aimed for and associated with a selected scenario. You can not set an objective on a "scenario" to be completed.

rgopal-sig (changes requested):
Hi Dennis, I think some clarification would benefit these notes. See my comments on respective files.

docs/reference/release-notes.md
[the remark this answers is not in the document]
dennis-sig: I copied most of your text, but I left out the comment about microservices. The trend towards microservices peaked around 2017-2020, what we're seeing now is a variety of system architectures, but all of them strive towards smaller systems.

docs/capabilities/reports/modernization-report.md
dennis-sig: Thanks, I updated the text accordingly.

rgopal-sig (approved):

PR #685, opened 2025-07-01
Clarify why certain risks do/don't apply to transitive dependencies.

dennis-sig:


dennis-sig:
@dessens0 We can't merge this yet until the actual change is rolled out, but what do you think about this explanation?

Issue #687, opened 2025-07-02
Check for scope configuration if system previously had a customer provided scope file


sbrinkhorst:
Sometimes a system had a customer provided scope file but for some reason this is no longer available.

We already retrieve the metadata of the system, which contains the information if the customer uses self-service scoping. If they do, check if the upload contains a `sigrid.y?aml` file and thrown an error if that is not the case.

PR #689, opened 2025-07-04
Block disappearing scope file

dennis-sig:
@sbrinkhorst As discussed. I moved the metadata around because we were loading it from the API multiple times.


test/test_sigridci_runner.py
```diff
@@ -653,6 +692,7 @@ def __init__(self, options, *, systemExists=True, uploadAttempts=0):
         self.POLL_INTERVAL = 1
 
     def callSigridAPI(self, path, body=None, contentType=None):
+        print(path)
```
sbrinkhorst, line 695: I guess you want to remove this again.

sigridci/sigridci/sigridci_runner.py
```diff
@@ -90,27 +91,32 @@ def prepareRun(self):
...
         scope = self.options.readScopeFile()
+
         if scope is not None:
             if self.options.subsystem not in (None, "", "root", "scopefile"):
                 UploadLog.log("Warning: You cannot provide a scope configuration file for a subsystem, it will be ignored.")
 
             self.validateConfiguration(lambda: self.apiClient.validateScopeFile(scope), "scope configuration file")
 
+        if scope is None and metadata.get("scopeFileInRepository"):
+            message = {"valid" : False, "notes" : ["Missing scope configuration file (sigrid.yaml)"]}
```
sbrinkhorst, line 117: I wonder if we need to make this clearer. 

"Missing scope configuration file (sigrid.yaml). Previous published analysis was based on custom scope file."

Not super happy about this proposal. Do you have better ideas? (If not we can skip it, we can always add it later if we have a good idea.)

sbrinkhorst (commented):
I made a ticket #687 , I don't know how to add it to the PR.

sigridci/sigridci/sigridci_runner.py
dennis-sig: How about this? I'm now linking to the documentation, which might be a nice way to elaborate on error messages. Do note the link doesn't work yet, since this same merge request is adding it.

sbrinkhorst: Yes, that looks good.

sbrinkhorst (approved):

PR #698, opened 2025-07-30
Update technology documentation.

dennis-sig:


docs/reference/technology-support.md
```diff
@@ -222,7 +222,6 @@ This page provides an overview of all technologies that can be analyzed by Sigri
 | `r`                           | R                                      | Maintainability, Architecture               |
 | `radience`                    | Radience                               | Maintainability, Architecture               | [(8)](#notes)     |
 | `razor`                       | Razor                                  | Maintainability, Architecture, AI           |
-| `react`                       | React                                  | Maintainability, Architecture, Security, AI | [(2)](#notes)     |
```
florislw, line 225: Maybe update the JS row with "JavaScript (includes React and NodeJS)" and the TS row with TypeScript (includes React). We can remove the includes Typescript remark, since that is obvious. Also, does Typescript also include NodeJS? If so, we might want to specify that as well.

dennis-sig: But does that actually help you choose? We also don't mention Angular by name, we only mention React and Vue by name because we used to have a separate analysis for it. People generally know whether they're using React "on top" of JavaScript or TypeScript, so they can still choose accordingly. What I'm trying to say is: This table doesn't give framework-specific advice anywhere else, so this is kind of an outlier.

florislw: Okay sure, we could add framework specific comments as needed, for example when support notices confusion.

PR #708, opened 2025-08-28
Exclude .so files

tukraus:
These are Linux shared libraries and not source code


dennis-sig:
I checked, and I can confirm that none of the 300+ languages we support uses .so for anything else.

dennis-sig (approved):

PR #827, opened 2026-03-27
Handle BitBucket not supporting line breaks in Markdown tables.

dennis-sig:


sigridci/sigridci/reports/report.py
```diff
@@ -61,6 +61,7 @@ def getSigridUrl(self, options):
 class MarkdownRenderer(ABC):
     def __init__(self):
         self.decorateLinks = True
+        self.tableLineSeparator = "<br />" if Platform.isHtmlMarkdownSupported() else " • "
```
dessens0, line 64: Where does Platform.isHtmlMarkdownSupported() come from?

dennis-sig: Ourselves:

```
    @staticmethod
    def isHtmlMarkdownSupported():
        if os.environ.get("SIGRID_CI_MARKDOWN_HTML") in ("false", "0"):
            return False
        return Platform.isGitHub() or Platform.isGitLab() or Platform.isAzureDevOps()
```

patveck (approved):

PR #855, opened 2026-04-20
Improvement recipes docs

florislw:


docs/integrations/sigrid-mcp/recipes.md
```diff
@@ -0,0 +1,89 @@
+# Sigrid Modernization Recipes MCP
```
bruno-oliveira, line 1: Should we say "Tool" instead of MCP here, since it's all one single MCP server?

bruno-oliveira (approved):

PR #861, opened 2026-04-28
Also (and separately) export merge commits as part of the history.

dennis-sig:


patveck (approved):
Looks good to me, no comments. The changes make sense, and the test runs the real thing, proving this works.

PR #862, opened 2026-04-29
Add image for system level ai governance page

Cheer-ding:


dennis-sig:
@Cheer-ding You need to make the file name lowercase, so starting with "system" instead of "System". The current name makes the pipeline fail.

bruno-oliveira (approved):
LGTM! 🚀 looking great!

PR #892, opened 2026-05-18
multi-repository-upload: support local folders, BOM handling, and temp dir options

frblondin:
### Local folder support (`sigrid-git-upload.py`)
Sources can now be either remote git URLs or local folder paths.
- Remote URLs are cloned as before.
- Local git repositories are exported via `git archive HEAD`, so only tracked files are included (node_modules, build outputs, etc. are excluded automatically).
- Non-git local folders are copied as-is.

### UTF-8 BOM handling
YAML files (`sigrid.yaml`, `sigrid-metadata.yaml`) are now read with `utf-8-sig` and written as plain UTF-8, stripping any BOM introduced by some editors/environments on Windows.

### `--keep-temp` flag
Prevents deletion of the temporary working directory after the run and prints its location, useful for debugging.

### `--temp-path` option
Allows specifying the parent directory for the temporary working folder. Useful on Windows when the default system temp path is long and combined with deeply nested repository paths would exceed the 260-character path limit.

### README updated
All new arguments and behaviours are documented.


maximjanssen:
I created a copy of this PR [here](https://github.com/Software-Improvement-Group/sigridci/pull/893) to enable organization-defined Sigrid feedback.

dennis-sig:
I merged Maxim's evolution in https://github.com/Software-Improvement-Group/sigridci/pull/893, so I'm closing this original one.

PR #895, opened 2026-05-19
Update VS Code extension docs with filtering and Jira integration

elnazmajlesiSIG:
Added documentation for two new features shipped in v0.0.7: filtering findings by risk and status, and creating Jira issues from findings. Also added new screenshots.


docs/integrations/vscode-extension.md
```diff
@@ -64,6 +66,66 @@ When opened, the Sigrid extension contains multiple tabs, one for each Sigrid ca
   default browser.
 - The pencil icon allows you to edit a finding's status and add remarks.
 
+## Filtering and searching findings
+
+When working with a large number of findings it can be hard to focus on what matters most. You can narrow
+down the findings list using the filter controls and the search bar at the top of the panel.
+
+**To filter by risk level**, click the filter icon (▽) next to the **Risk** column header. A dropdown appears
+with the following options:
```
andrtolv: Just keep in mind that these options are dynamic. It displays only options available for the active finding table. For example if there are no findings with status "Medium" the filter dropdown will not contain "Medium" option.

elnazmajlesiSIG: I've added it.

andrtolv (approved):

andrtolv:
Rename image files and replace instances to be in lower case. Because now unit tests are failing because of that:
`Image file name should be lowercase`

andrtolv (approved):

PR #904, opened 2026-05-29
add guardrails boyscout prompt

florislw:


docs/integrations/sigrid-mcp/guardrails.md
```diff
@@ -35,21 +33,26 @@ Add this to your agent instructions (see [where to place these instructions](#wh
...
 ## MANDATORY: Quality Gate
 
 Before reporting ANY task as complete:
 
-1. Run the Sigrid Code Quality Guardrails tool on all changed production code
-2. Maintainability findings: accept if principles were followed, otherwise refactor
-3. Security findings: fix if straightforward, otherwise flag to user
+1. Run the Sigrid Code Quality Guardrails tool on all files you changed
+2. Maintainability findings: fix every finding in files you touched, new or
+   pre-existing, judged against the principles above. Leave one only if the code
```
bruno-oliveira, line 45: Do we need this second sentence: "Leave one only if the code already honors the principles, or the fix cascades outside task scope (don't get stuck). Say which, and why."

Maybe we can just remove it? Or have you observed it getting stuck without it?

bruno-oliveira (approved):

docs/integrations/sigrid-mcp/guardrails.md
florislw: I have tested this prompt internally. Maybe we indeed can remove it (probably), but let's leave it for the next iteration now so we don't need to re-evaluate

PR #907, opened 2026-06-01
Fix CICD Feedback For On Premise

fonteijnf:
- See also internal Sigrid-Multi-Analyzer MR # 97
  - Added Cert passthrough to SigridCI feedback scripts
  - Added a single CI/CD token var for passthrough to SigridCI
- Added a GitHub feedback handler (instead of using actions, which might not work on-premises)
  - And a test
- Update docs, describing feedback behaviour for Sigrid-Multi-Analyzer
- Fixed a tiny bug for Azure feedback


sigridci/sigridci/feedback_provider.py
```diff
@@ -115,7 +116,8 @@ def prepareAdditionalReports(self, markdownReport):
             markdownReport,
             GitLabPullRequestReport(markdownReport),
             AzurePullRequestReport(markdownReport),
-            BitBucketPullRequestReport(markdownReport)
+            BitBucketPullRequestReport(markdownReport),
+            GitHubPullRequestReport(markdownReport),
```
dennis-sig, line 120: This would mean you get double GitHub feedback, one via the GitHub action and one from our own script. 

Is there a reason we can't use the existing GitHub action in the on-premise flow, necessitating our own implementation? If so, we need to make sure this *only* runs on-premise. At this stage we have so many people using the GitHub action that I don't think it's realistic we can force them to migrate due to this change, so we want this to be backward compatible.

dennis-sig (changes requested):

sigridci/sigridci/feedback_provider.py
fonteijnf: - Download: mshick is fetched from github.com at runtime, that's fine on SaaS, but it breaks on air-gapped GitHub Enterprise unless the action is mirrored. The new handler ships in the image.
- Cert: mshick has no cert option; the handler takes `SIGRID_GITHUB_CA_CERT_PATH`.
- No double comments: existing mshick users are untouched (their YAML is in their own repo). The handler doesn't get called unless `SIGRIDCI_GITHUB_COMMENT_TOKEN` is set. Double comment only if someone keeps mshick and sets the token.

I think we might even remove any mention of mshick and mention the GitHub token in docs.
The handler covers every mshick case and more, documenting both invites double-comment config.
mshick keeps working silently for those who have it.

dennis-sig: For on-premise that's true. So that would mean the following:

- Remove the mention of the separate action from the Sigrid CI for GitHub documentation.
- Add the documentation on how to configure the token, similar to how we do it for GitLab etc.
- See it working in action on this repository.
- The existing ones will continue to work, because they don't have the token (which is a clever solution).

sigridci/sigridci/reports/github_pull_request_report.py
```diff
@@ -0,0 +1,93 @@
...
+        request.add_header("X-GitHub-Api-Version", "2022-11-28")
+
+        api = ApiCaller("GitHub", pollInterval=5)
+        return api.retryRequest(lambda: urllib.request.urlopen(request, context=self.sslContext))
+
+    def getPullRequestNumber(self):
+        # GITHUB_REF format for pull requests: refs/pull/{number}/merge
+        parts = os.environ["GITHUB_REF"].split("/")
+        if len(parts) < 3 or not parts[2].isdigit():
+            raise ValueError(f"Could not extract PR number from GITHUB_REF: {os.environ['GITHUB_REF']}")
```
dennis-sig: I would not hard crash in this case. What you can do: In `isWithinGitHubPullRequestPipeline`, make sure this environment variable is available in the desired format.

dennis-sig (approved):

PR #932, opened 2026-06-18
Encode SCM branch/path data in PR comment links (F2)

haiyunxu:
## What

PR-comment Markdown links were built by interpolating SCM environment variables — the branch name and repo/project path — directly into the link target. On a pull request **from a fork**, the branch name is attacker-controlled and can contain characters such as `)`, `[`, `]`, or spaces that break out of the `[label](url)` Markdown and inject arbitrary content into the bot-authored comment.

This addresses finding **F2** from a recent security review (content/link spoofing in PR comments; Low severity but no downside to fixing).

## Changes

- **`platform.py`** — URL-encode the dynamic path and query segments in `createPullRequestFileURL` (GitLab/GitHub/Azure). `/` is preserved in path segments so legitimate slash-containing branches (`feature/x`) and file paths still resolve; query parameters fully encode `/`.
- **`reports/report.py`** — escape the Markdown link label so neither side of the `[label](url)` can be escaped.
- **`test/test_platform.py`** — add tests confirming normal slash-containing branches/paths are preserved and a malicious branch name is encoded.

## Testing

Full suite passes (199 passed; the only failures are unrelated: a pre-existing time-based roadmap-image check and a missing optional `jsonschema` dev dependency).

### DIFF
--- a/sigridci/sigridci/platform.py
+++ b/sigridci/sigridci/platform.py
@@ -14,6 +14,7 @@
 
 import os
 import sys
+import urllib.parse
 
 
 DOCS_URL = f"https://docs.sigrid-says.com"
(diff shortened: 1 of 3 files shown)


sigridci/sigridci/platform.py
```diff
@@ -70,25 +71,33 @@ def isHtmlMarkdownSupported():
 
     @staticmethod
     def createPullRequestFileURL(file, line=0):
+        # SCM-supplied values such as the branch name are untrusted (a fork PR
```
dennis-sig: Interestingly, the code totally works (I tested on all environments). But the comment is wrong, that's not what it does.That threw me off a bit, so I recommend we just drop this comment and leave the code.

dennis-sig (changes requested):

sigridci/sigridci/platform.py
haiyunxu: Good to know this. I will do that.

dennis-sig (approved):

PR #938, opened 2026-06-24
Always include scope file.

dennis-sig:


erik-blokland (approved):
(Test) changes look good.

PR #944, opened 2026-07-01
Add —detaillevel option

dennis-sig:
Sigrid CI will normally limit the feedback to prevent it from becoming overwhelming. @dhingrakriti has a customer that asks if we can add an option to show every single finding. This is not the default behavior, but this at least gives those people the option.


CLAUDE.md
```diff
@@ -0,0 +1,40 @@
...
+
+- Run unit tests: `python3 -m unittest`
+
+## Portability
+
+Sigrid CI is used across a variety of enterprise environments. Some people use the Docker container, other people
+run the code natively. This makes portability the top concern:
+
+- Do not update the required Python version. We need to be very conservative to ensure portability.
+- Do not depend on libraries in production code.
```
patveck:
```suggestion
- Do not depend on any external library in production code. Limit yourself to Python's standard library ONLY. 
```

patveck (approved):

PR #952, opened 2026-07-06
Docs/update jetbrains marketplace

elnazmajlesiSIG:


dennis-sig:
@elnazmajlesiSIG I think you can also remove the "beta" label in menu.html. Looks great otherwise!

dennis-sig (approved):

PR #954, opened 2026-07-07
Mention JetBrains integration in the release notes.

dennis-sig:


elnazmajlesiSIG (commented):
This is not a fair reflection of the situation, so this has been fixed so that only the available Git history is analyzed, and the missing period is simply considered not available.etter to write
to 
This does not accurately reflect the situation, so it has been corrected to only analyze the available Git history, treating the missing period as simply unavailable.

elnazmajlesiSIG (approved):

PR #959, opened 2026-07-14
Document Mendix Pipelines integration.

dennis-sig:


docs/sigridci-integration/mendix-pipelines.md
```diff
@@ -0,0 +1,87 @@
...
+|-------|-------|
+| **Base URL Path** | `https://sigrid-says.com/rest/inboundresults/qsm/{customer}/{system}` |
+| **Header 1 Key** | `Authorization` |
+| **Header 1 Value** | `$SigridToken`, or whatever you named your variable in step 1 |
+
+Then click **Save and activate**.
+
+## Verifying the integration
+
+Run your pipeline. A successful POST request step means the scan has been triggered, not that it has finished. Once the analysis completes, open your system in Sigrid and check that the analysis date reflects the current run.
```
maximjanssen: Do we want to point to the fact that Security and OSH show the timestamp of the latest analysis? That's better than pointing to the date IMO

maximjanssen (changes requested):

docs/sigridci-integration/mendix-pipelines.md
dennis-sig: Or just write something like "it might take a few minutes before you see the new results appear in Sigrid". There is nothing we can do about this, Mendix Pipelines only allows us to call an end point, so there is no way to be notified when the results are actually there.

maximjanssen (approved):

PR #971, opened 2026-07-23
Why do I see a message about Node.js 20 being deprecated in my pull request feedback

MichielCuijpers:
### Why do I see a message about Node.js 20 being deprecated in my pull request feedback?

Sigrid CI itself does not use Node.js, so this message does not come from Sigrid. If you're using the [GitHub Marketplace action](../sigridci-integration/github-actions.md#alternative-2b-github-marketplace) to post Sigrid CI feedback as a pull request comment, that action relies on a third-party GitHub Action ([`mshick/add-pr-comment`](https://github.com/mshick/add-pr-comment)) to post the comment. GitHub Actions is moving its runners from Node.js 20 to Node.js 24, and GitHub automatically adds this deprecation notice whenever a step still depends on the older Node.js runtime. Any accompanying Node.js `DeprecationWarning` lines (for example about `punycode` or `url.parse()`) come from that same dependency chain.

In short: this is a notice from GitHub Actions' infrastructure and the third-party comment action, not from Sigrid. It's a warning, not an error, and does not affect the Sigrid CI analysis itself. If your pipeline is actually failing, look for a separate error message in the same step's log, since the Node.js deprecation notice on its own will not fail the build.

### DIFF
--- a/docs/capabilities/faq.md
+++ b/docs/capabilities/faq.md
@@ -193,9 +193,9 @@ This message occurs when the (Linux) user that cloned your Git repository is dif
 
 ### Why do I see a message about Node.js 20 being deprecated in my pull request feedback?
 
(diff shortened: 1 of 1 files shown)


MichielCuijpers:
@dennis-sig I updated the PR

dennis-sig (approved):

PR #987, opened 2026-08-17
Add agents.py entrypoint to print structured feedback to stdout

florislw:
## Summary
- Adds `sigridci/agents.py`, a dedicated entrypoint (alongside `sigridci.py` and `sigridci_feedback.py`) that runs Sigrid CI in feedback-only mode and prints structured, JSON feedback to stdout (one capability-name line + one compact JSON line per capability) for AI agents to consume, instead of writing report files (`feedback.md`, `osh-feedback.md`, `security-feedback.md`, raw JSON dumps).
- `agents.py` always runs in `FEEDBACK_ONLY` mode with onboarding disabled and no `--out` option, since it never publishes or writes report files. `--disable-onboarding` is removed from `sigridci.py` and `action.yaml` since it's now only relevant to `agents.py` (always on there).
- Shared CLI logic (capability parsing, common argparse flags, system/subsystem name validation, and the run-and-exit flow) is extracted from `sigridci.py` into `sigridci/sigridci/cli_common.py` so `sigridci.py` and `agents.py` each only define the flags specific to them.
- Maintainability payload keeps close to the raw feedback but: drops legacy ratings not part of the current quality model (`VOLUME`, `COMPONENT_BALANCE_PROP`), drops the unused `refactoringCandidatesPerType` field, drops `occurrences` from refactoring candidates (redundant with `subject`), and replaces the five separate ratings dicts with a single `ratings` table (`systemBaseline`, `systemNew`, `changedCodeBefore`, `newAndChangedCodeAfter` per metric).
- OSH and security payloads reuse the existing `CycloneDXProcessor`/`SarifProcessor` decoders (same ones the markdown reports use) to emit only the vulnerability/license-risk-relevant libraries and clean finding objects, instead of dumping the raw CycloneDX SBOM / SARIF blob.
- Implements sig/development/sigrid-backlog#3985.

## Test plan
- [x] `python3 -m unittest` (217/217, one pre-existing unrelated failure: `jsonschema` not installed)
- [x] Verified payload shapes against real fixtures (`test/testdata/maintainability.json`, `osh-junit.json`, `security-sigrid-api-sarif.json`)
- [x] Ran a real `sigridci/agents.py` CLI invocation against a live system and inspected the stdout output


sigridci/sigridci/reports/inline_results_report.py
```diff
@@ -0,0 +1,126 @@
...
+        pass
+
+
+class MaintainabilityInlineResultsReport(InlineResultsReport, MaintainabilityMarkdownReport):
+    # Ratings that are still tracked in older Sigrid systems but are no longer part of
+    # the current quality model. See Objective.SYSTEM_PROPERTIES for the current metrics.
+    CURRENT_METRICS = Objective.SYSTEM_PROPERTIES + ["MAINTAINABILITY"]
+
+    # Maps the raw feedback's ratings dicts to the column names used in the "ratings" table below.
+    RATINGS_KEYS_NAMES = {
```
florislw, line 41: @dennis-sig can you or @Geertien confirm these "improved" keys from the raw json are correct. e.g. "newCodeRatings" we should call "newAndChangedCodeAfter"?
I want to ensure that what it is is not interpreted by agents.

sigridci/agents.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python3
```
dennis-sig, line 1: Nitpick, but I would call this `sigridci_agents.py` for symmetry with the other entry points.

florislw: Makes sense, but I actually did not do it on purpose. We are using this script for (quick) feedback anywhere in the loop. Meaning, this is not CI-related at all. It should be clear that this is not some special kind of CI to use when you also have agents somewhere. Maybe a better name would be `change_feedback_for_agents.py`. I will also rename the /ci-feedback skill to /change-feedback or something like that. What do you think?

sigridci/sigridci/cli_common.py
```diff
@@ -0,0 +1,78 @@
...
+
+def parseCapabilities(names):
+    try:
+        return [CAPABILITIES[name.lower().strip()] for name in names.split(",")]
+    except KeyError as e:
+        print(f"Invalid value for --capability: {str(e)}")
+        sys.exit(1)
+
+
+def addCommonArguments(parser):
```
dennis-sig, line 39: I do see the use for this utility file, but let's make it `cli_utils.py` or whatever and avoid the word "common". Because these options are not actually common, especially if we also consider the on-premise use case. I guess calling this the analysis options or something along those lines would be more descriptive. If you want to be fancy about it, you can also check the actual common options between the 3 entry points and put those in a function.

sigridci/sigridci/cli_common.py
```diff
@@ -0,0 +1,78 @@
...
+              f", not completely numeric, and be {PublishOptions.SYSTEM_NAME_LENGTH.start} to {maxNameLength} characters long (inclusive).")
+        sys.exit(1)
+
+    if not options.isValidSubSystemName():
+        print(f"Invalid subsystem name, subsystem name should match '{PublishOptions.SUBSYSTEM_NAME_PATTERN.pattern}'"
+              ", must be at least two characters long and not contain consecutive dots or slashes.")
+        sys.exit(1)
+
+
+def runSigridCi(options):
```
dennis-sig, line 66: See earlier comment about on-premise, maybe make this `runAnalysis` or something like that.

action.yaml
```diff
@@ -24,10 +24,6 @@ inputs:
     description: "Publishes your source code to Sigrid, without waiting for feedback (true/false)."
     required: false
     default: false
-  disableonboarding:
```
dennis-sig, line 27: Did you remove this because it's now implicit via the new entry point?

sigridci/sigridci/reports/inline_results_report.py
```diff
@@ -0,0 +1,126 @@
...
+from .osh_markdown_report import OpenSourceHealthMarkdownReport
+from .security_markdown_report import SecurityMarkdownReport
+from ..objective import Objective
+
+
+class InlineResultsReport(ABC):
+    def generate(self, analysisId, feedback, options):
+        payload = self.buildPayload(feedback, options)
+        print("Inline results: " + payload["capability"])
+        print(json.dumps(payload))
```
dennis-sig, line 28: Is it a problem that you're mixing this output with the normal logging in `stdout`? Or is that fine?

sigridci/sigridci/reports/inline_results_report.py
```diff
@@ -0,0 +1,126 @@
...
+    def buildPayload(self, feedback, options):
+        findings = self.extractFindings(feedback)
+        return {
+            "capability": self.getCapability().shortName,
+            "objectiveMet": self.isObjectiveSuccess(feedback, options),
+            "baseline": feedback.get("baseline"),
+            "findings": [self.buildFinding(finding) for finding in findings]
+        }
+
+    def buildFinding(self, finding):
```
dennis-sig, line 117: Assuming you want to export the entire thing, would `dataclasses.asdict(finding)` work?

dennis-sig (changes requested):

sigridci/sigridci/feedback_provider.py
```diff
@@ -111,6 +119,18 @@ def prepareMarkdownReport(self):
         else:
             raise Exception(f"Unknown capability: {self.capability}")
 
+    def prepareInlineResultsReport(self):
```
dennis-sig, line 122: This part duplicates `prepareMarkdownReport` a bit. In general it's not ideal we need to post-process the objectives received from the API in this way. But it's not like we add new capabilities every week, so it's fine like this.

action.yaml
florislw: yes

sigridci/sigridci/reports/inline_results_report.py
florislw: Its the goal, in the sense that now the agent only needs to call the script and read the output, no separate file lookup. If we later want less verbosity we can consider using quite/verbose flags

sigridci/sigridci/cli_common.py
florislw: Renamed to `cli_options.py` and split into `addSigridConnectionArguments` (the options actually shared by all three entry points, now used by `sigridci_feedback.py` too) and `addPublishArguments` (publish/CI-specific). Also moved `runSigridCi`/`validateOptions` out into `sigridci_runner.py` since they run the analysis rather than parse CLI options, renaming `runSigridCi` to `runAnalysis` in the process (see the other comment below). Pushed in f38aa21f.

sigridci/sigridci/cli_common.py
florislw: Moved to `sigridci_runner.py` and renamed to `runAnalysis`, alongside the connection-arguments split described in the comment above. Pushed in f38aa21f.

sigridci/sigridci/reports/inline_results_report.py
florislw: - finding.status.value won't work with asdict, we would need to do something like `{**dataclasses.asdict(finding), "status": finding.status.value}`
- this is an explicit step where we say, for a finding, we only want to display these fields to agents. If we use asdict, we lose this explicit step. So if a new value is added, we get it automatically. It is hard to predict (for me right now at least) whether that is something we would want or not.

florislw: Close, but not quite: `Finding` is a dataclass with exactly these 7 fields, so `dataclasses.asdict(finding)` reproduces most of this. However `status` is an `Enum`, and `asdict()` doesn't unwrap enum members, so `json.dumps()` right after this in `generate()` would fail with `TypeError: Object of type FindingStatus is not JSON serializable`. We'd still need to special-case `status.value` on top of `asdict()`. Given that, keeping the explicit dict here for now.

sigridci/sigridci/feedback_provider.py
florislw: Agreed, but handling this duplication is outside of the scope of this MR then maybe?

sigridci/sigridci/reports/inline_results_report.py
dennis-sig: Sorry, I missed your comment but you are indeed correct. The display names are different because we took multiple attempts at trying to communicate what everything means.

sigridci/sigridci/feedback_provider.py
dennis-sig: OK. I can take care of it when adding the architecture feedback, since that will duplicate things even more.

sigridci/agents.py
dennis-sig: I see the reasoning, but we're in a repository called Sigrid CI :) In reality this is now more about real-time feedback in terms of use case, which may occur outside of a CI environment. I'll merge it like this for now, this file existing in a directory called `sigridci` should make it explicit enough what it's for.

dennis-sig (approved):
I'm assuming we do *not* want release notes for this yet, as this is currently not yet used as a flow.

florislw:
> I'm assuming we do _not_ want release notes for this yet, as this is currently not yet used as a flow.

yes. We will have to update the skill to use this flow first.

PR #999, opened 2026-08-31
Enable SIgrid CI for Security by default and remove Beta label

dennis-sig:


sigridci/sigridci/cli_options.py
```diff
@@ -19,7 +19,7 @@
 
 
 CAPABILITIES = {cap.shortName: cap for cap in [MAINTAINABILITY, OPEN_SOURCE_HEALTH, SECURITY]}
-DEFAULT_CAPABILITIES = "maintainability,osh"
+DEFAULT_CAPABILITIES = "maintainability,osh,security"
```
maximjanssen, line 22: With the increase in the amount of capabilities and the upcoming CI for AQ launch I'm kinda worried that customers are accidentally locking themselves out of AQ feedback when they just want to disable e.g. security. Would it be worth considering to add an alternative option `--disable-capability`?

dennis-sig: Good suggestion, that's also more in line with what people want to do (i.e. run everything *except* X or Y).

docs/sigridci-integration/using-sigridci.md
```diff
@@ -122,12 +122,17 @@ you can explicitly add `--capability maintainability` to *only* receive feedback
...
-### Security feedback (Beta)
+### Security feedback
 
 Sigrid CI provides security feedback based on your [objectives](../capabilities/portfolio-objectives.md).
 This helps you to "shift left" by reporting security findings early in the development process, making it easier
 and faster to address those findings. 
 
+If your Sigrid license includes Sigrid Security, you will automatically receive security feedback in Sigrid CI.
+If you prefer *not* to receive security feedback, you can use
+[the `--capability` option](../reference/client-script-usage.md#command-line-options) to disable it.
```
maximjanssen: In this case I'd point to the --disable-capability option

docs/sigridci-integration/using-sigridci.md
maximjanssen, line 1: Can you also update the "Adding Open Source Health feedback to an existing Sigrid CI configuration" section to use `--disable-capability` instead of the reference it includes now to `--capability maintainability` (which also disables security)?

maximjanssen (changes requested):

maximjanssen (approved):

